PR #16, opened 2023-12-30
Fix UP/Down/Left/Right/Home/End when APP_CURSOR mode is enabled

### LINKED ISSUE #13: Arrow keys not working in `less`
Not sure why, given it works in bash and nvim, `sys.stdin.readline()` in python shows the exact same bytes for arrow keys as xterm, and the `TERM` value isn't the issue.

hjkl however works fine.

MoSal:
Fixes #13, shell history, and many more.

 All this code is probably ripe for some refactors and higher
 abstractions. But this is good enough for now.

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -3,6 +3,7 @@
 use alacritty_terminal::{
     index::{Column as TermColumn, Point as TermPoint, Side as TermSide},
     selection::{Selection, SelectionType},
+    term::TermMode,
 };
 use cosmic::{
     iced::{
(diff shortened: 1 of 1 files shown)


MoSal:
To test this, add a `dbg!(is_app_cursor);` at the start of the `Up` case:

* In zsh, type a partial command in zsh, and hit the `Up` key. `APP_CURSOR` mode should be `true`. Testing with `less` also works.
* Run `cat` without arguments, then hit the `Up` key. In this case, `APP_CURSOR` is `false`, and you can see that the sequence passed is not the `APP_CURSOR` one.

jackpot51 (approved):

Issue #14, opened 2023-12-27
Feature request: zoom


leviport:
Zoom in: `Ctrl` + `+` (some terminals require `shift` so it's a real `+` instead of a `=`)
Zoom out: `Ctrl` + `-`
Zoom 100%: `Ctrl` + `0`

PR #22, opened 2024-01-03
Implement ZoomIn/ZoomOut/ZoomReset with configurable step

MoSal:
Implements #14.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -31,6 +31,8 @@ pub struct Config {
     pub app_theme: AppTheme,
     pub font_name: String,
     pub font_size: u16,
+    pub font_size_zoom_adj: i8,
+    pub font_size_zoom_step_mul_100: u16,
     pub show_headerbar: bool,
     pub syntax_theme_dark: String,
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -42,6 +44,8 @@ impl Default for Config {
             app_theme: AppTheme::System,
             font_name: "Fira Mono".to_string(),
             font_size: 14,
+            font_size_zoom_adj: 0,
```
jackpot51, line 47: Should this default be 0?

jackpot51: Looks like 0 is correct, based on the other code

jackpot51:
This looks good, so I merged. I'm not sure if I want to save the zoom to config and, in doing so, sync it across instances. It seems to me like permanent adjustments ought to be done by changing the font size, while zoom can be used for temporary adjustments that are per-tab.

leviport:
In my experience, zoom resets when closed in most terminal apps. Gnome Console is the only one I can recall off the top of my head that remembers zoom levels. So I think that sounds good as well.

PR #23, opened 2024-01-03
Move zoom_adj to App so it doesn't persist between runs

MoSal:
Addressing comments from #22.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -31,7 +31,6 @@ pub struct Config {
     pub app_theme: AppTheme,
     pub font_name: String,
     pub font_size: u16,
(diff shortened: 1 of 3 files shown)


jackpot51:
I think it needs to be removed from Config, so it does not sync between instances. I can do the change if you like.

MoSal:
> I think it needs to be removed from Config, so it does not sync between instances. I can do the change if you like.

I changed the commit moving. `zoom_adj` to `App`.

I imagined this would have needed to be a bigger change for some reason.

jackpot51 (approved):

PR #27, opened 2024-01-04
Configurable font stretch, font weight, and bold font weight

MoSal:
* Store a font name => font faces map

 * Only list font names that have `NORMAL` and `BOLD` faces with
   `Normal` stretch. This will give us defaults and fall-backs that are
   guaranteed to always exist.

 * Filter by stretch first, with `Normal` chosen as the always existing
   default.

 * Then only list font weights supported by the font name stretch
   selected, for the normal and bold cases.

 * When changing the font name selected, the stretch/weight options
   will stay the same if supported by the new font name, or revert to
   the always existing fall-backs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -874,6 +874,7 @@ dependencies = [
  "lazy_static",
  "libcosmic",
  "log",
+ "paste",
  "rust-embed",
  "serde",
  "tokio",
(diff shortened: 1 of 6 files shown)


MoSal:
**Open Question:** Do font weights/stretches need to be translatable?

jackpot51:
This works very well, but due to the large number of font settings, I wonder if they should be organized in a different section.

MoSal:
@jackpot51 Is the commit I just pushed similar to what you had in mind?

jackpot51:
This does look good now. There are conflicts, please resolve them, and I can merge.

jackpot51 (approved):

PR #33, opened 2024-01-08
Bright bold hidden and no show cursor

MoSal:
* Use bright color if bold.
* Support `HIDDEN` cell flag.
* Hide cursor if `TermMode::SHOW_CURSOR` is not set

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -461,6 +461,10 @@ impl Terminal {
                         )
                     };
 
+                    if indexed.cell.flags.contains(Flags::HIDDEN) {
+                        fg = bg;
+                    }
+
                     // Change color if cursor
                     if indexed.point == grid.cursor.point {
                         //TODO: better handling of cursor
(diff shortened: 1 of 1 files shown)


MoSal:
**Note:** Regarding `BOLD`+`fg=Foreground` and dark themes. Other terminals (not `alacritty`) may set this to a color that is actually dimmer than normal. This is definitely easier on the eyes in bold-heavy content.

Should this be considered a theme concern (probably the sensible choice), or should this be special-cased?

src/terminal.rs
```diff
@@ -436,22 +443,36 @@ impl Terminal {
 
                     let mut attrs = self.default_attrs;
 
+                    let cell_fg = if indexed.cell.flags.contains(Flags::BOLD) {
+                        as_bright(indexed.cell.fg)
```
jackpot51, line 447: I wouldn't do this by default, since there is already an ANSI escape code to use the bright colors, and bold already sets a different font.

MoSal:
Apologies for the accidental push which included the dim changes (fixed).

Anyway, rebased, and bright bold is a disabled-by-default toggle now.

jackpot51 (approved):

jackpot51:
Thanks!

PR #37, opened 2024-01-10
Fix double tab creation at startup, and another minor change

MoSal:
* Sleep for 50ms before sending first `Message::TabNew`.
* Use `log::warn!()` instead of `println!()` when reporting non-handled term events (see commit message).


jackpot51 (approved):

jackpot51:
Thanks!

PR #38, opened 2024-01-10
Add LEFT‑TO‑RIGHT ISOLATE character to the beginning of lines

MoSal:
This zero-width char will force the shaper to treat detected RTL
 lines as LTR.

 RTL text would still be rendered correctly. But this fixes the wrong
 behavior of it being displayed aligned to the right.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -414,6 +414,12 @@ impl Terminal {
     }
 
     pub fn update(&mut self) -> bool {
+        // LEFT‑TO‑RIGHT ISOLATE character.
+        // This will be added to the beginning of lines to force the shaper to treat detected RTL
+        // lines as LTR. RTL text would still be rendered correctly. But this fixes the wrong
+        // behavior of it being aligned to the right.
+        const LRI:char = '\u{2066}';
+
         let instant = Instant::now();
 
         //TODO: is redraw needed after all events?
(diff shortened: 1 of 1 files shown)


jackpot51:
What is wrong about right aligned RTL text?

MoSal:
> What is wrong about right aligned RTL text?

Nothing.. outside terminals.

I'll give a few examples below which are just meant to demonstrate some practical concerns and showcase some issues. But note that RTL behavior in terminals is relatively well defined nowadays, and it's meant to behave like how `cosmic-term` would behave with this change. You can try a VTE-based terminal to compare.

```
echo 'hello\nمرحبا\nhello\nمرحبا'
```

Beyond the eye travel, try to select that Arabic word.

Let's do something else:

```
echo -e 'عادي \033[7mمعكوس\033[0m عادي'
echo -e 'a عادي \033[7mمعكوس\033[0m عادي'
```

Oops. Setting bg on cells is broken in the first line, but not in the second.

Alright, let's demonstrate a practical case:

```
mkdir دفتر1
touch دفتر1/file1
touch دفتر1/file2
mkdir دفتر2
touch دفتر2/file1
touch دفتر2/file2
ls -l دفتر*
```

More eye travel.

MoSal:
An option to let the shaper auto-detect direction can be added later (maybe with keyboard shortcut too). That would indeed make a nice addition for RTL wall-of-text use-cases.

jackpot51:
@MoSal are you a native Arabic speaker? It would be nice to have an Arabic translation so I can ensure RTL is working in libcosmic and iced.

MoSal:
> @MoSal are you a native Arabic speaker?

Yes. But I don't use translated interfaces, which makes me someone who is not that familiar with the technical terminology often used. So I can't competently contribute translations, I'm afraid.

I can help review/double-check contributions from others, if that's ever needed.

PR #60, opened 2024-01-17
Support splitting of terminals using a Pane Grid

snaggen:
I have implemented split terminals adding a pane grid. 
The keys are the same as Tilix
Ctrl-Alt-R  for split vertical
Ctrl-Alt-D for split horizontal
Ctrl-Shift-X for toggle maximised. 
Ctrl-Shift-<arrow> to navigate between panes

Outstanding things to be clarified and fixed:
1. In terminal_box.rs I have blocked crtl-alt control keys from being sent, since ctrl-alt-d would cause an exit if not. Not sure if this is safe, or if it only should prevent this for the specific key bindings. 
2. I have added a shade to the background, by just making it slightly darker for non focused panes. This should probably come from the theme, where a non-focused background could be specified or something like that. How this should be handled properly must be decided by some that knows more about the cosmic desktop.

### DIFF
--- a/i18n/en/cosmic_term.ftl
+++ b/i18n/en/cosmic_term.ftl
@@ -48,3 +48,6 @@ menu-settings = Settings...
 
 # Context menu
 show-headerbar = Show header bar
+split-horizontal = Split Horizontal
+split-vertical = Split Vertical
+pane-toggle-maximize = Toggle Pane Maximized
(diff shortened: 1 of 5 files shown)


snaggen:
My take on shading the background was far to naive... so that should probably be done properly. However, that code was quite hairy, so I would appreciate some assistance here.

snaggen:
I have now cleaned it up and synced with master so feel free to review this.

jackpot51 (approved):

PR #70, opened 2024-01-19
Handle Ctrl-Shift-u to input unicode

snaggen:
I have created a small state machine in terminal_box. By pressing Ctrl-shift-u, the input gets in to Unicode state, after that, all ascii hex digits are stored in a Vec. Then on any non-ascii hex digit or modifier change, the stored digits are converted to the unicode character they represent and sent to the terminal.


snaggen:
I have updated this to master, it is now much less intrusive. The only thing for this feature would be to show some special unicode indication at the cursor (possibly 'Ⓤ' ) when you are in unicode input mode. However, the cursor handling still looks quite limited, so I'm not sure how to do this. Now, all terminals doesn't have this, so I think it is still usable without. 
Ctrl-Shift-u is hardcoded, but this is the standard used for unicode input under linux, so I don't think that is a problem.

snaggen:
Maybe this should be handled higher up in the input stack though.

feikedonia:
This issue is also present in Cosmic text editor

Issue #49, opened 2024-01-12
Copy / Paste using PRIMARY


snaggen:
I tested to select text with the mouse and then tried to paste it using the middle mouse button, but with no success. So, I assume that this might be something that just is not implemented yet. 

I tested it under a Gnome desktop (latest Fedora) , if that is relevant.

MoSal:
Funny, I thought I would be the only one who would miss this.

I like `PRIMARY` in terminals because  `Shift+Insert` is easier than `Ctrl+Shift+v`.

Anyway, I think support for this needs to be added to [window_clipboard](https://github.com/hecrj/window_clipboard) first, then `iced`.

I was planning to look into it at some point if no one beats me to it.

jackpot51:
As @MoSal said we need changes to window_clipboard. I'm also working on changes for cosmic-files that would add custom mime types.

gantryyork:
https://github.com/pop-os/cosmic-term/issues/55
These two could be considered the same issue

jackpot51:
This should be implemented with the middle mouse button

PR #72, opened 2024-01-20
Add support for Primary

snaggen:
Fixing: https://github.com/pop-os/cosmic-term/issues/49 

I have added support for the Primary clipboard. Note that this depends on https://github.com/hecrj/window_clipboard/pull/26  and https://github.com/pop-os/iced/pull/95 (and then libcosmic must use that iced version of course).


snaggen:
I have the copy and middle click paste handled in terminal_box, the rationale for that is that I thought that it was a fundamental feature of the terminal. But, now when I have tought about it a while, I start to wonder if I instead should have on_middle_click and on_selected methods, that would allow for all the clip board handling to be moved in to main.rs and handled much the same way as regular copy/paste. 

What do you think?

MoSal:
FWIW, that's how I implemented it:
https://gist.github.com/MoSal/8f74535219db54aba76bc193b1a962f6

Didn't add middle click paste. But it would be just like Shift+Insert.

snaggen:
I can adapt this PR to this. Feel much cleaner.

snaggen:
To handle middle click, I first need to update focus to the pane we click on. So I had to use an on_middle_click method for that. So I added a MiddleClick message to handle this. Then I have CopyPrimary and PastePrimary since that will allow for conditional compiling, if cosmic-term can build on platforms without primary.

MoSal:
> Then I have CopyPrimary and PastePrimary since that will allow for conditional compiling, if cosmic-term can build on platforms without primary.

Oh, the way I added primary support to `window_clipboard` and `iced` does not require this:
https://github.com/hecrj/window_clipboard/pull/25
https://gist.github.com/MoSal/8f74535219db54aba76bc193b1a962f6#file-iced-patch

So this will depend on what upstream dependencies will go with.

But do note that `#[cfg(target_family = "unix")]` wouldn't work. Check the top of `lib.rs` in `window_clipboard` to see why.

snaggen:
But will your window_clipboard PR compile on anything other than Linux? 
It looks like you didn't have conditional compilation on the trait, and I didn't see any implementation of the primary traits for other platforms?

Now, I just looked at it on my phone, so I may have missed things . Phones are not optimal for looking at patches. 😊

MoSal:
> But will your window_clipboard PR compile on anything other than Linux?

Yes. The trait has default impls returning `None`, which turns into no-ops in iced.

The iced side already uses `Option`s. So what's `clipboard.read().ok()` for CLIPBOARD is simply `clipboard.read_primary().and_then(|res| res.ok())` for PRIMARY, where `read_primary()` would return `None` if platform does not implement it.

snaggen:
Should I update my PRs, or do you want to go with yours?

MoSal:
> Should I update my PRs

Go ahead.

snaggen:
Ok, closed my window_clipboard PR, moved my iced PR to use https://github.com/hecrj/window_clipboard/pull/25 and removed all conditional compilation since that is no longer needed.

snaggen:
When https://github.com/iced-rs/iced/pull/2240 gets merged, and libcosmic is updated to include that patch, then we can fix this PR

snaggen:
When https://github.com/iced-rs/iced/pull/2240 is now merged, so we are one step closer now. Just libcosmic left to start using that before we can merge this.

snaggen:
Seems we are getting closer. From what I understand, the relevant clipboard patches have reached libcosmic. Unfortunately there still seems to be some version conflicts related to bitflags. But it looks like we are getting closer.

snaggen:
Bumping alacrity_terminal to 0.23.0 seems to resolve the bitflags (but needs other fixes). But I was unfortanately unable to compile this since the dependencies have been to complex for me to just swap out libcosmic to a local version.

snaggen:
Took the time to try to get things to compile yesterday, but I'm confused about libcosmic. It seems it uses iced 0.12, which have the primary patch, but then there are quiet a lot of changes on top of that which breaks this patch. And libcosmic seems to no longer have a read/write_primary for every matching clipboard read/write. So, I feel that I will not fiddle around with these changes since I don't understand the purpose of them or where to even submit PRs (since these changes doesn't seem to exist in upstream iced).

jackpot51:
The iced that libcosmic uses has two major changes that are not yet in upstream iced:

- iced_sctk is a renderer specifically for wayland desktops that allows for the creation of special surfaces like layer shell and lock surfaces
- accesskit is integrated for accessibility

If you give me more information about what is missing or changed from what you expect, I can help to bring the necessary changes into our iced fork.

snaggen:
This pull in libcosmic should add back the needed api
https://github.com/pop-os/iced/pull/127

PR #75, opened 2024-01-22
Focus Follows Mouse

snaggen:
I have added an option to let the focus follow the mouse pointer, without having to click. It is a sloppy focus, so it only changes focus on mouse pointer enter. 

Two things: 
1. I know that Focus Follows Mouse sounds like a window manager level thing, but since we have panes, we need to support it on the application level, since focus between panes also follows the mouse. 
2. I also realize that this may fall under weird workflows (ref: https://xkcd.com/1172/ ), but I hope you find it useful enough to include anyway  :)  I was kind enought to hide it behind a setting that defaults to off.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,10 +24,10 @@ palette = "0.7"
 git = "https://github.com/pop-os/cosmic-text.git"
 
 [dependencies.libcosmic]
-git = "https://github.com/pop-os/libcosmic.git"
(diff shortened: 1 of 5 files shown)


jackpot51:
I think this makes sense. We may adjust some of these features as the rest of the team gets a look.

jackpot51 (approved):

PR #81, opened 2024-01-25
A simple password manager

snaggen:
I have implemented a simple password manager, which uses libsecret to store and read passwords.

Note, that this is my first experience with iced, so beware that the context_drawer page is far from optimal. I had trouble finding a good list that would allow me to select available passwords. So, it is currently a list of buttons, but it seems that buttons will not fill a row.... but it looks ok and works.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ i18n-embed-fl = "0.6"
 rust-embed = "6"
 paste = "1.0"
 palette = "0.7"
+libsecret = { version = "0.4.0", features = ["v0_20"] }
+gio = "0.18.4"
 
 [dependencies.cosmic-text]
 git = "https://github.com/pop-os/cosmic-text.git"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,7 @@ i18n-embed-fl = "0.6"
 rust-embed = "6"
 paste = "1.0"
 palette = "0.7"
-libsecret = { version = "0.4.0", features = ["v0_20"] }
-gio = "0.18.4"
+secret-service = { version = "3.0.1", features = ["rt-tokio-crypto-rust"] }
 
 [dependencies.cosmic-text]
 git = "https://github.com/pop-os/cosmic-text.git"
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,6 +50,7 @@ mod terminal;
 use terminal_box::terminal_box;
 mod terminal_box;
 
+mod password_manager;
 mod terminal_theme;
 
 lazy_static::lazy_static! {
@@ -176,6 +177,7 @@ pub enum Action {
     Paste,
     SelectAll,
     Settings,
+    PasswordManager,
     ShowHeaderBar(bool),
     TabActivate0,
     TabActivate1,
@@ -212,6 +214,7 @@ impl Action {
             Action::Paste => Message::Paste(entity_opt),
             Action::SelectAll => Message::SelectAll(entity_opt),
             Action::Settings => Message::ToggleContextPage(ContextPage::Settings),
+            Action::PasswordManager => Message::ToggleContextPage(ContextPage::PasswordManager),
             Action::ShowHeaderBar(show_headerbar) => Message::ShowHeaderBar(show_headerbar),
             Action::TabActivate0 => Message::TabActivateJump(0),
             Action::TabActivate1 => Message::TabActivateJump(1),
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -19,6 +19,8 @@ i18n-embed-fl = "0.6"
 rust-embed = "6"
 paste = "1.0"
 palette = "0.7"
+libsecret = { version = "0.4.0", features = ["v0_20"] }
+gio = "0.18.4"
```
mmstick, line 23: We shouldn't depend on GNOME libraries.

snaggen: I looked around a bit for alternatives, but it seems that there are not many good options available. Do you have any alternatives you prefer?

snaggen: Does cosmic desktop have any current or planned keyring?

mmstick: A crate will need to be developed that provides zbus bindings for https://specifications.freedesktop.org/secret-service/latest/.

snaggen: Well, that expands the scope of this password manager quite a lot 😅

snaggen: I see that zbus is using async_io::block_on for its blocking wrappers. So, would it be an ok solution to use the async secret-service library with a similar blocking wrappers?

mmstick: See https://github.com/pop-os/dbus-settings-bindings for existing zbus bindings we have.

mmstick: There is no need. zbus lets you choose a runtime, and libcosmic has a tokio feature.

mmstick: https://crates.io/crates/secret-service
This has a `rt-tokio-crypto-rust` feature.

snaggen: I just saw that they provide a blocking API also, so I will probably not have to explicitly depend on tokio....

mmstick: libcosmic already uses tokio for its runtime, so it would make sense to use it.

snaggen: I have now replaced libsecret with the secret-service crate.

src/password_manager.rs
```diff
@@ -0,0 +1,244 @@
...
+            self.cached_password_list = Some(self.get_password_list());
+        }
+    }
+
+    pub fn clear_password_cache(&mut self) {
+        self.cached_password_list = None;
+    }
+
+    pub fn get_password(&self, identifier: &str) -> Option<String> {
+        let ss = match secret_service::blocking::SecretService::connect(
```
mmstick, line 31: This should be spawned with a `Command` with the async version so that it doesn't block the UI.

snaggen: Ok, as I said I'm new to iced, so this is a bit unfamiliar territory for me. But I will look in to this.

mmstick: You can return a `Command` from the `update` method of the application. We have a convenience method for creating commands that spawn futures, which is `cosmic::command::future()`.

src/main.rs
```diff
@@ -263,6 +266,11 @@ pub enum Message {
     MouseEnter(pane_grid::Pane),
     Paste(Option<segmented_button::Entity>),
     PasteValue(Option<segmented_button::Entity>, String),
+    Password(String),
```
mmstick, line 269: It'd be safer if passwords were stored with https://crates.io/crates/secstr

snaggen: Well, benefit of using secret service, is that is is normally unlocked when the user is logged in. Using other alternatives means you have to enter a password to unlock it. And since the use of passwords in the terminal is fairly rare, that would mean that you would have to unlock this almost every time you would use it, which would cancel out any benefits.

snaggen: Deleted my previous comment, since I misunderstood....

But, yes, that sounds like a nice improvement. Will put that on the list.

jackpot51 (commented):
I don't agree with including this functionality in a terminal.

snaggen:
> I don't agree with including this functionality in a terminal.

Why? In the phone and web we use password mangers to allow us to have long secure passwords. I admit, I use ssh-keys for most things in the terminal these days, but there are still quite a few passwords to enter a little now and then. Most of the time, I end up opening up Bitwarden to copy and paste them, or just having bad passwords. 
Having a builtin password managers will make it easier to have secure passwords for things that we still need passwords for.

mmstick:
You could make an application like this: https://gitlab.gnome.org/World/secrets

src/main.rs
snaggen: Looking at the code now, that is not actually a password, but the identifier to the password that is retreived and sent to the terminal.

snaggen:
Well, we already have separate applications, like bitwarden. But that is quite distruptive, to start up a different application, search for the password you need, click copy to clipboard. Then go back to the terminal and paste. 
Ctrl-Alt-p and click is much less disruptive.

snaggen:
I realize that my last two comments can come out as a bit rude, just want to clarify that that was not at all my intention. I'm not a native English speaker so sometimes (especially when in a hurry) things come out short, and less polite than intended. I apologize if you perceived my comments in this way.

I fully understand if a password manager doesn't fit your vision for Cosmic Term, even if I personally find it convenient. 
I just came here to thank @mmstick for the reviews, I have now updated my branch with the suggested improvements so it is now in a much better shape. I will keep this branch around for my personal use and try to keep it up to date with the main branch unless it becomes to cumbersome. So, if you change your mind in the future, it will hopefully be available.

PR #88, opened 2024-01-30
Mouse Events

### LINKED ISSUE #8: Mouse emulation
Support sending mouse events to programs

snaggen:
I have made a first attempt at implementing mouse events. I have tested this with vim and w3m in SGR mode, which seems to work. However, if someone knows how to test normal/utf8 mode let me know, since that really needs testing. I have looked at some other implementations and read the spec and things are confusing. It seems that terminals encode the utf8 differently and some other differences. I have tried to follow the spec as much as I can, but it is a bit unclear around some things. 

Fixes #8

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,6 +32,7 @@ use tokio::sync::mpsc;
 
 use config::{AppTheme, Config, CONFIG_VERSION};
 mod config;
+mod mouse_reporter;
 
 use icon_cache::IconCache;
 mod icon_cache;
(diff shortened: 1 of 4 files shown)


snaggen:
Also, cursor movements should probably be sent.

snaggen:
Added cursor movement for SGR mode. So, that works pretty well now.

Now, it is just normal/utf8 mode left, but for that I need to figure out how to test it.

snaggen:
I have gone over the normal/utf8 mode and tested the calculations against other implementations. So, I think this is ready for review now.

jackpot51 (approved):
Thanks!

PR #95, opened 2024-01-31
Add some common syntax themes

MoSal:
Add themes (or palette/scheme pairs as they're referred to elsewhere)
 commonly found and used in other terminals:

 * Tango Palette (Dark and Light themes)
 * XTerm Palette (Dark and Light themes)
 * Rxvt Palette (Dark and Light themes)
 * Solarized Palette (Dark and Light themes)
 * Linux Console Palette & Theme

-------

If these are too many, at least the XTerm ones should be added, IMHO, as they are common White-On-Black and Black-On-White themes. `XTerm Dark` is also what I personally use, FWIW.

### DIFF
--- a/src/terminal_theme.rs
+++ b/src/terminal_theme.rs
@@ -138,6 +138,313 @@ fn auto_colors() -> Colors {
     colors
 }
 
+fn tango_palette() -> Colors {
+    let mut colors = auto_colors();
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):
Thank you!

Issue #50, opened 2024-01-12
New instances sometimes launch with multiple tabs


XV-02:
When launching a new instance of Cosmic-Term in either Gnome or Cosmic in Pop!_OS, the application launches with two tabs open. 

One tab is titled "New Terminal" and the other takes the format of "`user`@`hostname`:~". 

This is true if launches from the launcher, using the `super + t` shortcut, from the application view, from the dock, or from a terminal.

When I close the "`user`@`hostname`:~" tab, the application remains open, and switches to the "New Terminal" tab - which is entirely non-responsive, so far as I can tell.

jackpot51:
Please test with master staging, this should be fixed by #37

XV-02:
Looks like this was remedied before I opened it! #37 successfully addresses this.

XV-02:
I'm going to reopen this issue, and tweak the title a little. 

While I don't see cosmic-term opening with multiple tabs every time, I am seeing it infrequently. 

More over, I can semi-reliably make it happen about once in ten times while I am aggressively spamming the `super + t` shortcut to launch terminals in a Cosmic DE session.

snaggen:
This should be fixed by https://github.com/pop-os/cosmic-term/commit/4401a7dbef3e1579af9694c5f6c38c1b788a86e2

PR #106, opened 2024-02-01
Moving first terminal creation

snaggen:
By moving the creation of the first terminal to when the terminal event channel is created, we can avoid to rely on sleeps at startup. This could possibly fix the issue with two terminal tabs #50. 

The downside with this patch is that it might be a bit weird to create the first terminal in the TermEventTx message handler, but then not sure if it was less weird to create it in the subscription initializer.


snaggen:
This broke things it seems when I did some more testing...

PR #117, opened 2024-02-05
chore: update libcosmic

wash2:
keyboard input has changed a lot in iced, but I think I've made the proper adjustments here. vertical splits don't seem to work, but that didn't seem to work before either. Also, there is a black border on the right and bottom of the first tab of the terminal. I'm not sure why yet, but it seems like it is probably related to the scrollbar

Requires: https://github.com/pop-os/libcosmic/pull/287

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,10 @@ rust-embed = "6"
 paste = "1.0"
 palette = "0.7"
 
+[dependencies.smol_str]
+version = "0.2.1"
+features = ["serde"]
+
 [dependencies.cosmic-text]
 git = "https://github.com/pop-os/cosmic-text.git"
 features = ["shape-run-cache"]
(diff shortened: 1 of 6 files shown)


wash2:
There are still black bars, and this needs a rebase, but it should be otherwise good.

Cargo.toml
```diff
@@ -21,8 +21,13 @@ rust-embed = "6"
 paste = "1.0"
 palette = "0.7"
 
+[dependencies.smol_str]
+version = "0.2.1"
+features = ["serde"]
+
 [dependencies.cosmic-text]
 git = "https://github.com/pop-os/cosmic-text.git"
+rev = "1b025ae"
```
jackpot51, line 30: I'd really prefer not to lock cosmic-text to a specific commit.

ids1024:
I wonder if relying on serializing Iced's `Key` type might not be ideal if we want a stable format for keybindings... It's awkward to deal with though. It could also make sense to use the same format for representing keybindings in both the compositor and applications.

But removing `Serialize`/`Deserialize` from `KeyBind` doesn't seem to break compilation, so I guess keybinds aren't currently configurable? So at least that doesn't break anything and doesn't block this update.

Issue #125, opened 2024-02-07
Some control keys are sent twice


snaggen:
This issue is not really dependent on the key refactoring https://github.com/pop-os/cosmic-term/pull/122 , but since that added support for more keys in general, some duplicates were also added. 

Ctrl-Delete sends both the escape code ^[[3;5~ and the control code ^? , this was an issue before the key rewrite also. The correct code for Ctrl-Delete is the escape code ^[[3;5~  , so it must be prevented that Ctrl-Delete is sent form https://github.com/pop-os/cosmic-term/blob/fec6738af5c935e135b2819ecba7fee40fbeec32/src/terminal_box.rs#L743 
Ctrl-Backspace, Ctrl-Enter and Ctrl-Tab gets sent twice, to fix them, the ctrl variant of these could be removed from the block at https://github.com/pop-os/cosmic-term/blob/fec6738af5c935e135b2819ecba7fee40fbeec32/src/terminal_box.rs#L667 
Still, that doesn't solve Ctrl-Delete. Not sure if there is some "Correct" solution that is prettier than just removing them from the block and adding a special case for Ctrl-Delete.

snaggen:
I started to look at handle this by simple exceptions, but that gets quite ugly and confusing since some keys like Ctrl-Tab, Ctrl-Delete must be handled in the key press handler, but then also gets handled in the CharacterRecieved handler. Then other like Ctrl-Backspace can just be solved by removing them from the key pressed handler, but that gets a confusing special case. So it is quite inconsistent the whole handle keys both in CharacterRecieved and KeyPressed.

snaggen:
Would it be possible to use a state to indicate that a character is already handled by the key press event? 
Like I the key press set the state, then clear it in the key release event?

PR #128, opened 2024-02-08
Prevent keys from being handled in both KeyPressed and CharacterRecieved

snaggen:
Some keys must be handled in KeyPressed, but are also getting handled in CharacterRecieved. This adds a flag that gets set in the key released event if the key gets handled there, and then cleared in the key released handler. That flag is then checked in the CharacterRecieved handler to avoid sending characters twice. 

Fixes: https://github.com/pop-os/cosmic-term/issues/125

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -661,6 +661,7 @@ where
                 };
                 if let Some(escape_code) = escape_code {
                     terminal.input_scroll(escape_code);
+                    state.handled_by_key_press_event = true;
                     return Status::Captured;
                 }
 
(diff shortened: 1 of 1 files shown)


snaggen:
Not sure this is really correct though.... you may get multiple key press events before any key release event. And event though I assume CharacterRecieved is close to the press event, there might be a risk of races. Now, this is a low risk, and it only affects the special handled keys, but not entirely happy but not sure how to best handle the issue #125 .

snaggen:
PR #129 might be a better solution, even though it is just adding a special case for the problematic keys.

Issue #135, opened 2024-02-12
Bound keys are still handled by terminal


snaggen:
There are key bindings defined in src/key_bind.rs, when you use these key bindings they are still handled by the terminal.
It seems like we handle Ctrl+Alt for C0 control keys according to https://sw.kovidgoyal.net/kitty/keyboard-protocol/#ctrl-mapping  , but then we have ctrl-alt-d bound to horizontal split. And it seems that zsh, for some reason unknown to me, have choosen to handle ctrl-alt-d the same as ctrl-d, causing zsh users to being unable to do horizontal split by keyboard (context menu works of course). 

So, even if the simple fix here could be to change the key binding for horizontal split, the underlying issue is still that keys bound by the UI should be caught and prevented from reaching the terminal.

PR #136, opened 2024-02-14
Prevent bound keys from reaching the terminal

snaggen:
Check if a key is bound by the UI, and prevent it from reaching the terminal. 
Fixes: #135


snaggen:
I notice that the fmt also modified src/main.rs.... well, consider that a bonus :)

PR #139, opened 2024-02-16
Remove double key prevention

snaggen:
After rewrite to use Named keys, the double key prevention code instead made Ctrl-Bakspace et al. to not work again. Now the prevention code is not needed any more, so this patch removes it.


jackpot51 (approved):

jackpot51:
Needs a rebase after #136

snaggen:
Rebase done

PR #144, opened 2024-02-24
Add ability to select a default profile.

leb-kuchen:
This enables to select a default shell without changing the system shell.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -322,6 +322,7 @@ pub enum Message {
     ZoomOut,
     ZoomReset,
     FocusFollowMouse(bool),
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -855,7 +856,14 @@ impl App {
                                     },
                                 ),
                             ),
-                        );
+                        )
+                        .add(widget::settings::item::builder("Make default").control(
```
jackpot51, line 860: This should be added to the english translation

jackpot51 (approved):
It looks good. We may adjust the UI at some point but this is a good feature to have.

PR #147, opened 2024-02-27
fix: improve font rendering on light theme w/ iced web-colors feature

mmstick:
Screenshot comparison below. Currently, it is difficult to read dark text on a light background with a non-HiDPI display. Iced gained a `web-colors` feature recently which makes text in this scenario significantly more legible as a workaround.

screenshot-2024-02-27-20-38-49


mmstick:
We could enable it by default in libcosmic. Perhaps it's not necessary to make it optional.

jackpot51 (approved):

PR #148, opened 2024-02-28
i18n(sr): Update translations

git-f0x:
Tried to reduce the number of commits, so that they don't fill up as much of history.
Unrelated: Is there a chance for the 24.04 installer to default to btrfs with transparent compression, since I believe it was discussed some time ago?


jackpot51 (approved):

jackpot51:
It is unlikely we will be switching to btrfs for 24.04 by default, but we may offer it as an option.

git-f0x:
Yeah, it would be great if there was an option to select a primary filesystem for automatic partitioning.
Also this PR turned out a bit messy with the commit referencing the first issue, so it's probably best that I do this normally in the future.

PR #160, opened 2024-03-14
Add command line argument to allow a no-fork mode

sixpounder:
This is a quick edit I had to make in order to debug a slight issue I had on my system, and I thought could be useful to other people as well.

I did _not_ add any fancy command line argument parser such as `clap`, just moved the command line argument parsing on top of the `main` function and allowed it to parse a `--no-daemon` argument to avoid process forking.

Forking is still the default behaviour.


Quackdoc:
I've run into the same issue on other cosmic applications. I was thinking about making PRs for pretty much the same thing on some of the other cosmic applications as a compile time feature flag

ChocolateLoverRaj (approved):
I tried it and it works. It's great for developing cosmic-term because I can just close the terminal and `cargo r` command with either Ctrl+C or closing the window.

jackpot51 (approved):

PR #176, opened 2024-04-10
Rename gruvbox color scheme to match all the others

snaggen:
It have always bothered me that the gruvbox color scheme didn't have the same naming convention that the other color schemes. This is just a simple rename to fix that.

### DIFF
--- a/color-schemes/Gruvbox Dark.ron
+++ b/color-schemes/Gruvbox Dark.ron
@@ -1,5 +1,5 @@
 (
-    name: "gruvbox-dark",
+    name: "Gruvbox Dark",
     foreground: "#EBDBB2",
     background: "#282828",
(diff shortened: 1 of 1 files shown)


gold:
> It have always bothered me that the gruvbox color scheme didn't have the same naming convention that the other color schemes. This is just a simple rename to fix that.

This change is straightforward, innocuous, and sensible.

Suggestion:

On a related note, the filename convention in the `color-themes/` directory is likewise inconsistent -- in the opposite way!

lower-case-no-spaces is preferred (hyphen or underscore word delimiters; I don't care which you choose). And only `gruvbox-dark.ron` is currently  in the preferred filename style.

gold (approved):
LGTM

with comment/suggestion

jackpot51 (approved):

PR #178, opened 2024-04-12
Add fallback if selected color theme is not present

snaggen:
The fallback is the first available theme when sorted, this is to make things predictable and to get the same theme if you open the terminal multiple times

I added this since the color theme handling changed, and I ended up with only a titlebar... thought this would be better.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,8 @@ use std::sync::OnceLock;
 use crate::fl;
 
 pub const CONFIG_VERSION: u64 = 1;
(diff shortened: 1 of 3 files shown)


jackpot51:
It would probably be better to fall back to the default COSMIC themes.

snaggen:
Ok, will try to fix that.

snaggen:
Now the COSMIC themes are used.

jackpot51 (approved):

Issue #179, opened 2024-04-13
[Profiles] Profile tabs with command line close immediately


git-f0x:
When a profile has a command line option (e.g. neofetch, lsblk), the tab using that profile is immediately closed.
If that profile is the default profile, the terminal closes immediately after launching.

leb-kuchen:
You may find this alacritty github issue helpful: https://github.com/alacritty/alacritty/issues/1672
For fish and nu you have use the init-command and execute flags respectively if you do not want to exit the shell. For other shells I do not know.

git-f0x:
This is with default bash in Pop!_OS.
If this is an issue in alacritty, and cannot be fixed in cosmic-term, then the command line option should probably be removed until it is fixed upstream, since it's not exactly expected behavior that using this option would just close the tab.

mmstick:
cosmic-term does not use the alacritty binary, so what applies there does not apply here. In general, a terminal will naturally exit once the profile's command has exited.

git-f0x:
So is this intended behavior? I assumed that Profiles are used more for aesthetic reasons (e.g. neofetch on launch, customized tab names) or to run something (script, etc.) and continue using the terminal.

leb-kuchen:
Shell options are provided by the alacritty_terminal crate, so the linked  issue is very relatable. 
https://docs.rs/alacritty_terminal/0.23.0/alacritty_terminal/tty/struct.Shell.html

leb-kuchen:
I use them to set a default shell and I do not want to run `chsh -s`, cause I am a little scared of running exotic shells in a tty.

leb-kuchen:
alacritty_terminal provides a hold option, so it could be implemented quiet easily. I do not know if it should be default, the advantage is that you still have access to rest of the GUI.

PR #181, opened 2024-04-14
Add tty working dir and hold

leb-kuchen:
See #179 for more reasons why these options are useful.
For the working directory, if the shell does not provide to execute a command at startup, you will need to something like `bash -c "cd somewhere; exex bash"`.
For hold, I think true should be the default , since like in the issue above, the terminal will exit if it also a  default, which means you will have to edit or remove the config file manually and have no GUI. 
Hold probably should be a drop down like in gnome terminal, for now just a showcase.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -291,7 +291,9 @@ pub enum Message {
     PasteValue(Option<segmented_button::Entity>, String),
     ProfileCollapse(ProfileId),
     ProfileCommand(ProfileId, String),
+    ProfileDirectory(ProfileId, String),
     ProfileExpand(ProfileId),
+    ProfileHold(ProfileId, bool),
     ProfileName(ProfileId, String),
     ProfileNew,
     ProfileOpen(ProfileId),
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -899,11 +911,29 @@ impl App {
...
                                 ),
                             ),
+                        )
+                        .add(
+                            widget::row::with_children(vec![
+                                widget::column::with_children(vec![
+                                    widget::text(fl!("hold")).into(),
+                                    widget::text::caption(fl!("remain-open")).into(),
+                                ])
+                                .spacing(2)
```
jackpot51, line 926: Is there a reason why the spacing is 2 here and space_xxxs is commented out?

jackpot51:
This looks good if that manual spacing item can be resolved.

leb-kuchen:
I did some experimenting and thought the  lowest spacing of 4 is too much.  Anyway does your UX team prefer a dropdown over a toggler, cause I think these options can be explained more cleanly?

jackpot51:
I think either a toggler or a dropdown is fine. The lowest spacing of 4 was chosen to ensure that fractional scaling in multiples of 25% always ends up with an integer.

jackpot51 (approved):

jackpot51:
This looks good enough to me now. We may adjust later based on UX feedback.

PR #195, opened 2024-05-02
Smart ctrl c

ChocolateLoverRaj:
I can rebase to not include my `nix` branch if needed. Closes #114


i18n/de/cosmic_term.ftl
```diff
@@ -49,6 +49,11 @@ default-dim-font-weight = Matte Schriftstärke
 default-bold-font-weight = Fette Schriftstärke
 use-bright-bold = Fetten Text heller darstellen
 
+### Behavior
+behavior = Behavior
```
jackpot51, line 53: You don't have to fill in every translation file. It is best to only fill in the languages you know, and it always falls back to the English variant if a language misses a translation. This makes it easy for people translating other languages to see what is missing.

.envrc
```diff
@@ -0,0 +1,2 @@
+use flake
```
jackpot51, line 1: I'd avoid including this change

jackpot51:
I'll need to talk with the team about this change.

i18n/de/cosmic_term.ftl
ChocolateLoverRaj: At first I only added the English file but I got an error with the pl! macro

jackpot51:
We are on board with this change, and do not need to have a setting for it. Please clean up the translation files, and remove the setting, so that Ctrl+C will always copy when text is selected.

ChocolateLoverRaj:
It's great that Smart Ctrl+C will be on by default. Are you sure that you don't want to let people turn it off if they don't want it though?

i18n/de/cosmic_term.ftl
git-f0x: Not sure if I'm doing it the wrong way, but it would be way easier to find what's missing if the changes were synced across files (since it often requires rearranging and careful checking, especially in large files).

jackpot51:
Yes, I'm sure.

src/main.rs
```diff
@@ -1669,6 +1677,8 @@ impl Application for App {
                         let term = terminal.term.lock();
                         if let Some(text) = term.selection_to_string() {
                             return Command::batch([clipboard::write(text), self.update_focus()]);
+                        } else {
+                            println!("Copied, but no text selection!");
```
ChocolateLoverRaj, line 1681: At this point how can a `^C` signal be sent to the terminal?

ChocolateLoverRaj:
I'm assuming that Ctrl+Shift+C will also copy. So Ctrl+C will be "smart copy" and Ctrl+Shift+C will always copy. So should there be another shortcut called "Smart Copy" which is set to Ctrl+C?

src/main.rs
jackpot51: Sending 0x03 with the input_scroll function.

ChocolateLoverRaj:
What about pasting with Ctrl+V? I personally would prefer if both Ctrl+V and Ctrl+Shift+V pasted. I'm not sure what `^V` does but whenever it appears it's always from me accidentally not holding the Shift key.

jackpot51:
Ctrl+V is for triggering verbatim mode. Due to legacy, we can't globally change the way these things work.

src/key_bind.rs
```diff
@@ -22,7 +22,7 @@ pub fn key_binds() -> HashMap<KeyBind, Action> {
 
     // Standard key bindings
     bind!([Ctrl, Shift], Key::Character("A".into()), SelectAll);
-    bind!([Ctrl, Shift], Key::Character("C".into()), Copy);
```
jackpot51, line 25: Please leave in Ctrl+Shift+C. You can do another binding for Ctrl+C like CopyOrCtrlC to make it clear.

ChocolateLoverRaj: Ideally, would the Copy function just have two shortcuts, Ctrl+C and Ctrl+Shift+C?

jackpot51: Two shortcuts, with different actions. Ctrl+Shift+C is Copy, Ctrl+C is CopyOrCtrlC. This way, Ctrl+Shift+C will always copy and never send a signal.

.envrc
ChocolateLoverRaj: Should I add this file to `.gitignore` in the #194 PR and rebase this PR?

src/key_bind.rs
ChocolateLoverRaj: I named it `CopyOrSigint` for less confusion.

PR #232, opened 2024-06-22
profile flag

leb-kuchen:
This supports the explicit flags `-PVALUE`  and `--profile=VALUE`.
Implicit flags are also easy to add.


wiiznokes:
Why not use clap at this point ?

leb-kuchen:
It is a huge dependency. I think [lexopt](https://docs.rs/lexopt/latest/lexopt/) would also work fine.

wiiznokes:
Yeah but clap have a good support for help messages and everything

jackpot51:
If this is rebased with the conflicts fixed, I can merge it.

jackpot51:
I'd prefer to use `clap` for this. We already use it on some other COSMIC applications.

atiq-cs:
Thank you for adding profile support @leb-kuchen

PR #233, opened 2024-06-26
fix: Use X-COSMIC as a category, not COSMIC

ryanabx:
This makes the desktop file "valid". When we run `desktop-file-validate`, we currently get:

```
cosmic-term.x86_64: E: invalid-desktopfile /usr/share/applications/com.system76.CosmicTerm.desktop value "COSMIC;System;TerminalEmulator;" for key "Categories" in group "Desktop Entry" contains an unregistered value "COSMIC"; values extending the format should start with "X-"
```

I'll be making PRs for the other COSMIC desktop files, too :)

### DIFF
--- a/res/com.system76.CosmicTerm.desktop
+++ b/res/com.system76.CosmicTerm.desktop
@@ -7,6 +7,6 @@ Terminal=false
 Type=Application
 StartupNotify=true
 Icon=com.system76.CosmicTerm
-Categories=COSMIC;System;TerminalEmulator;
+Categories=X-COSMIC;System;TerminalEmulator;
 Keywords=Command;Shell;Terminal;
 Keywords[pl]=Cli;WierszPoleceń;Powłoka;Terminal;


leb-kuchen:
Why though, Gnome and Kde are valid. Just do nothing and it will become valid at some point.

wash2 (approved):
I think it is ok to prefix the value with `X-` until COSMIC becomes a category. :+1:

jackpot51:
I'd rather see a change to make COSMIC considered valid.

ryanabx:
> I'd rather see a change to make COSMIC considered valid.

This would be preferable long-term.

Downstream we are holding package reviews for cosmic and this is one of the blockers for acceptance

mmstick:
It'd be better to patch it if you require it. It won't be necessary for packaging in Pop!_OS.

ryanabx:
> It'd be better to patch it if you require it. It won't be necessary for packaging in Pop!_OS.

That's what I'll probably end up doing 😁

PR #263, opened 2024-07-27
fix ctrl or alt + arrows

### LINKED ISSUE #149: Alt or Ctrl + left/right arrows behaves incorrectly
Alt or Ctrl + left/right arrows should move the cursor forward or backward 1 word. Currently, it moves the cursor forward or backward by one character and inserts unwanted characters. Ctrl + arrows inserts `;5` and alt + arrows inserts `;3`

To recreate: 
- In a terminal, type a few words separated by spaces
- Hold alt or ctrl, then tap the left arrow

Version: `0.1.0~1709152706~22.04~36477e0`

### LINKED ISSUE #253: Implement Alt+Left/Right from Gnome terminal
These should mirror Alt+B and Alt+F which jump backwards and forward one word.

### LINKED ISSUE #260: Can't use zellij `Alt + <⬅⬆⬇➡>` shorcuts
When using [zellij](https://zellij.dev/) I can't navigate the terminal panes with `Alt + <⬅⬆⬇➡>`; instead `":3"` are inputted in the prompt.

image

This does not render zellij unusable in "cosmic-term" since zellij also allows `Alt + <hjkl>`, which does work.

leb-kuchen:
should fix #149, should fix #253, and should fix #260

Since the refactor in #122, csi sequences have been used for arrow sequences if `is_app_cursor` was set , before ss3 had been used for ctrl and alt modifiers. 
The custom logic for shift keys was applied to all modifiers in this commit. 
Even though I do not know, if this logic is still need, I still preserved it for shift keys. 
The  correct modifier for shift is 2 and was not applied before the patch, thus it could had been a workaround.

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -600,7 +600,6 @@ where
 
         let is_app_cursor = terminal.term.lock().mode().contains(TermMode::APP_CURSOR);
         let is_mouse_mode = terminal.term.lock().mode().intersects(TermMode::MOUSE_MODE);
-
         let mut status = Status::Ignored;
         match event {
             Event::Keyboard(KeyEvent::KeyPressed {
@@ -613,6 +612,7 @@ where
                         return Status::Captured;
                     }
                 }
+
                 let mod_no = calculate_modifier_number(state);
                 let escape_code = match named {
                     Named::Insert => csi("2", "~", mod_no),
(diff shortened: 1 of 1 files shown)


leb-kuchen:
https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-functional-keys
The 2. form becomes the 3. form if no modifiers are set, so the `is_app_cusor` logic is correct,  but csi should not have been used.

leb-kuchen:
@tobyrg Which shell are you using? I can select text in fish.
With ` kitty +kitten show_key -m kitty` you can see it correctly sends the text

tobyrg:
@leb-kuchen I'm using the default bash shell on Pop!_OS.

leb-kuchen:
Are you on the lattest commit?

tobyrg:
Yes I'm on the latest commit, Ctrl Arrows key works great for moving arrow words but not selecting text. This is what I get from `kitty +kitten show_key -m kitty` 

```
Shift+LEFT PRESS
CSI 1 ; 2 D

Shift+RIGHT PRESS
CSI 1 ; 2 C
```

leb-kuchen:
Which shell?

tobyrg:
Bash, but I don't think is actually a thing in Bash from what I've found online.

drakon64:
Patch c72010062db45f2d5a4655fa47ebd9f2617c30a4 is working fine against be808b56cf24d03fc99cf44b0885078a81a16523

jackpot51 (approved):

PR #271, opened 2024-08-03
i18n(en-GB): Add British English localisation

RossBarnie:


i18n/en-GB/cosmic_term.ftl
```diff
@@ -0,0 +1,99 @@
...
+## Settings
+settings = Settings
+
+### Appearance
+appearance = Appearance
+theme = Theme
+match-desktop = Match desktop
+dark = Dark
+light = Light
+syntax-dark = Color scheme dark
```
jackpot51, line 39: Do you want these to be colour?

RossBarnie: Yes, it would have helped if I had actually committed the changes I was wanting to make! Fixed now.

jackpot51 (approved):

PR #280, opened 2024-08-10
Make underlines stable

### LINKED ISSUE #277: [BUG] Undercurl/dotted/dashed jitters when moving cursor
When moving the cursor over an under-curl as seen in the video, it "jitters", as in the curling moves around.

Applies to:

- undercurl
- dashed underline
- dotted underline

[Screencast from 2024-08-08 23-55-04.webm](https://github.com/user-attachments/assets/ca16dd3e-9608-47f2-a35e-1b005bed138d)

A-Walrus:
Closes #277

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -406,71 +406,69 @@ where
                                 renderer.fill_quad(underline2_quad, line_color);
                             }
 
+                            let mut draw_repeated = |rects: &[(f32, Option<f32>)]| {
+                                let full_width = self.end_x - self.start_x;
+                                let pattern_len: f32 = rects.iter().map(|x| x.0).sum();
+                                let mut pos = 0.0;
+                                let mut index = {
+                                    let in_pattern = (self.start_x) % pattern_len;
+
+                                    let mut sum = 0.0;
+                                    let mut index = 0;
+                                    for (i, rect) in rects.iter().enumerate() {
+                                        sum += rect.0;
+                                        if in_pattern < sum {
+                                            let width = sum - in_pattern;
+                                            if let Some(height) = rect.1 {
+                                                let pos_offset = mk_pos_offset!(pos, height);
+                                                let underline_quad =
+                                                    mk_quad!(pos_offset, style_line_height, width);
+                                                renderer.fill_quad(underline_quad, line_color);
+                                            }
+                                            index = i + 1;
+                                            pos += width;
+                                            break;
+                                        }
+                                    }
+                                    index
+                                };
+                                while pos < full_width {
+                                    let (width, x) = rects[index % rects.len()];
+                                    if let Some(height) = x {
+                                        let pos_offset = mk_pos_offset!(pos, height);
+                                        let underline_quad =
+                                            mk_quad!(pos_offset, style_line_height, width);
+                                        renderer.fill_quad(underline_quad, line_color);
+                                    }
+                                    pos += width;
+                                    index += 1;
+                                }
+                            };
+
                             if metadata.flags.contains(Flags::DOTTED_UNDERLINE) {
                                 let bottom_offset = style_line_height * 2.0;
-
-                                let full_width = self.end_x - self.start_x;
-                                let mut accu_width = -(self.start_x % 4.0);
-                                let mut dot_width = 2.0f32.min(full_width - accu_width);
-
-                                while accu_width < full_width {
-                                    dot_width = dot_width.min(full_width - accu_width);
-                                    let pos_offset = mk_pos_offset!(accu_width, bottom_offset);
-                                    let underline_quad =
-                                        mk_quad!(pos_offset, style_line_height, dot_width);
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += 2.0 * dot_width;
-                                }
+                                draw_repeated(&[(2.0, Some(bottom_offset)), (3.0, None)]);
                             }
 
                             if metadata.flags.contains(Flags::DASHED_UNDERLINE) {
                                 let bottom_offset = style_line_height * 2.0;
-
-                                let full_width = self.end_x - self.start_x;
-                                let dash_width = 6.0f32;
-                                let gap_width = dash_width / 2.0;
-
-                                let mut accu_width = -(self.start_x % (gap_width + dash_width));
-
-                                while accu_width < full_width {
-                                    let width = dash_width.min(full_width - accu_width);
-
-                                    let pos_offset =
-                                        mk_pos_offset!(accu_width.max(0.0), bottom_offset);
-                                    let underline_quad = mk_quad!(
-                                        pos_offset,
-                                        style_line_height,
-                                        width + (accu_width).min(0.0)
-                                    );
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += width + gap_width;
-                                }
+                                draw_repeated(&[(6.0, Some(bottom_offset)), (3.0, None)]);
                             }
 
                             if metadata.flags.contains(Flags::UNDERCURL) {
                                 let style_line_height = style_line_height.floor();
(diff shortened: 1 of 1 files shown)


src/terminal_box.rs
```diff
@@ -410,7 +410,7 @@ where
                                 let bottom_offset = style_line_height * 2.0;
 
                                 let full_width = self.end_x - self.start_x;
-                                let mut accu_width = 0.0;
+                                let mut accu_width = -(self.start_x % 4.0);
                                 let mut dot_width = 2.0f32.min(full_width - accu_width);
```
MoSal, line 414: With negative `accu_width`, the `.min()` call would not limit the width of the underline to the span.

Same with the other styles below.

A-Walrus: I reworked them and unified into one function for drawing repeating pattern underlines. I also make sure thatit starts and stops within the span :+1:

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
                                 renderer.fill_quad(underline2_quad, line_color);
                             }
 
-                            if metadata.flags.contains(Flags::DOTTED_UNDERLINE) {
-                                let bottom_offset = style_line_height * 2.0;
-
+                            let mut draw_repeated = |rects: &[(f32, Option<f32>)]| {
```
MoSal, line 409: Maybe add a comment describing the meaning of the tuple fields, or use a struct, although maybe that's an overkill.

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
-                                    dot_width = dot_width.min(full_width - accu_width);
-                                    let pos_offset = mk_pos_offset!(accu_width, bottom_offset);
-                                    let underline_quad =
-                                        mk_quad!(pos_offset, style_line_height, dot_width);
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += 2.0 * dot_width;
+                                let pattern_len: f32 = rects.iter().map(|x| x.0).sum();
+                                let mut pos = 0.0;
+                                let mut index = {
+                                    let in_pattern = (self.start_x) % pattern_len;
```
MoSal, line 414: Useless parenthesis!

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
-                                while accu_width < full_width {
-                                    dot_width = dot_width.min(full_width - accu_width);
-                                    let pos_offset = mk_pos_offset!(accu_width, bottom_offset);
-                                    let underline_quad =
-                                        mk_quad!(pos_offset, style_line_height, dot_width);
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += 2.0 * dot_width;
+                                let pattern_len: f32 = rects.iter().map(|x| x.0).sum();
+                                let mut pos = 0.0;
+                                let mut index = {
```
MoSal, line 413: A comment or two in this block might also be helpful.

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
+                                            }
+                                            index = i + 1;
+                                            pos += width;
+                                            break;
+                                        }
+                                    }
+                                    index
+                                };
+                                while pos < full_width {
+                                    let (width, x) = rects[index % rects.len()];
```
MoSal, line 436: You forgot to limit the width here. Easy to spot with `./ansi-colors.sh` and dashed underlines ;)

``` rust
let (mut width, x) = rects[index % rects.len()];
width = width.min(full_width - pos);
```

Also `pos` is not very descriptive IMHO. Maybe going back `accu_width` or something similar would be preferred. Feel free to rename `full_width` too if you feel like it.

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
+                                        renderer.fill_quad(underline_quad, line_color);
+                                    }
+                                    pos += width;
+                                    index += 1;
                                 }
+                            };
+
+                            if metadata.flags.contains(Flags::DOTTED_UNDERLINE) {
+                                let bottom_offset = style_line_height * 2.0;
+                                draw_repeated(&[(2.0, Some(bottom_offset)), (3.0, None)]);
```
MoSal, line 450: Is that a copy error, or do you prefer `3.0` width gaps over `2.0` ones ;)

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
-
-                                while accu_width < full_width {
-                                    dash_width = dash_width.min(full_width - accu_width);
-                                    let pos_offset = mk_pos_offset!(accu_width, bottom_offset);
-                                    let underline_quad =
-                                        mk_quad!(pos_offset, style_line_height, dash_width);
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += dash_width + gap_width;
-                                }
+                                draw_repeated(&[(6.0, Some(bottom_offset)), (3.0, None)]);
```
MoSal, line 455: Add a comment describing the tuples passed and/or bring back the descriptive variable names (`dash`, `gap`).

src/terminal_box.rs
```diff
@@ -406,72 +406,69 @@ where
...
 
-                                let full_width = self.end_x - self.start_x;
-                                let mut accu_width = 0.0;
-                                let mut dot_width = 1.0f32.min(full_width - accu_width);
-
-                                while accu_width < full_width {
-                                    dot_width = dot_width.min(full_width - accu_width);
-
-                                    let dot_bottom_offset = match accu_width as u32 % 8 {
+                                let pattern: Vec<_> = (0..8)
```
MoSal, line 462: Let's not use the heap for this.

``` rust
// add array to std `use` statements
let pattern: [(f32, Option<f32>); 8] = array::from_fn(|i| match i {
    3..=5 => (1.0, Some(bottom_offset + style_line_height)),
    2 | 6 => (1.0, Some(bottom_offset + 2.0 * style_line_height / 3.0)),
    1 | 7 => (1.0, Some(bottom_offset + 1.0 * style_line_height / 3.0)),
    _ => (1.0, Some(bottom_offset)),
});

```

MoSal (changes requested):
This is almost ready, but not quite.

Left a few comments.

src/terminal_box.rs
```diff
@@ -406,72 +407,79 @@ where
...
-
-                                while accu_width < full_width {
-                                    dash_width = dash_width.min(full_width - accu_width);
-                                    let pos_offset = mk_pos_offset!(accu_width, bottom_offset);
-                                    let underline_quad =
-                                        mk_quad!(pos_offset, style_line_height, dash_width);
-                                    renderer.fill_quad(underline_quad, line_color);
-                                    accu_width += dash_width + gap_width;
-                                }
+                                let dot = (2.0, Some(bottom_offset));
```
MoSal, line 461: `dash` and `6.0`.

src/terminal_box.rs
```diff
@@ -406,72 +407,79 @@ where
...
+                                        3..=5 => (1.0, Some(bottom_offset + style_line_height)),
+                                        2 | 6 => (
+                                            1.0,
+                                            Some(bottom_offset + 2.0 * style_line_height / 3.0),
+                                        ),
+                                        1 | 7 => (
+                                            1.0,
+                                            Some(bottom_offset + 1.0 * style_line_height / 3.0),
+                                        ),
+                                        _ => (1.0, Some(bottom_offset)),
```
MoSal, line 480: nit.

Should have mentioned this earlier, but I think it's better to be more explicit here now and use:

``` rust
  0 => (1.0, Some(bottom_offset)),
  _ => unreachable!(),
```

MoSal (changes requested):
This should be ready after fixing dashed underlines, and addressing a nitpick.

MoSal (approved):

jackpot51 (approved):
Thanks for this work @A-Walrus, and thank you for the reviews @MoSal!

PR #293, opened 2024-08-17
Only update default colors when needed, and `update()` when it's done

### LINKED ISSUE #291: [Bug] Resetting text size (Ctrl+0) while already at default text size crashes terminal
If you use Ctrl+0 without changing the text size, or after changing it and then changing it back to normal text size crashes the terminal window.

MoSal:
With commit 7c5d5440e5a560238de3666d7ead6970477a7dd6, it became
 possible that `update_colors()` may run without being immediately
 followed by an `update()`. But `update_colors()` creates a new
 `metadata_set`, which means `attr_list` may now contain out-of-date
 metadata indices to the old cleared `metadata_set`.

With this commit:
     * A `metadata_set` is only cleared and recreated in `update_colors()`
       if default colors have actually changed.
     * A `bool` is returned and set it to `true` from `update_colors()` if
       a color update happened.
     * An `update()` is run if we get a `true` from `update_colors()`.
     * And finally, `update_colors()` is renamed to
       `update_default_colors()`.

 Fixes #291.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -580,31 +580,39 @@ impl Terminal {
             }
         }
 
-        //TODO: this is done on every set_config because the changed boolean above does not capture
+        // NOTE: this is done on every set_config because the changed boolean above does not capture
         // WINDOW_BG changes
-        self.update_colors(config);
+        let default_colors_updated = self.update_default_colors(config);
 
         if update_cell_size {
             self.update_cell_size();
-        } else if update {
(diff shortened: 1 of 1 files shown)


MoSal:
Just to note, I don't see the panic from #291, and I don't know the issue 7c5d5440e5a560238de3666d7ead6970477a7dd6 fixed. But I think the fix here is logically sound.

jackpot51 (approved):

Issue #227, opened 2024-06-17
Changing text size causes cosmic-term to freeze/lag


tobyrg:
Using the shortcuts or manually changing the text size cosmic-term will hang/freeze/lag for a second.

GrownPlanet:
This seems to happen every time `App::save_config` is called.
this are some results of my rather primitive testing:
```
App::update_config took 0 ms // happened at start-up
// happens when I zoom in/ out, change font size or font, ...
App::update_config took 2 ms
App::update_config took 11 ms
App::update_config took 5 ms
App::update_config took 5 ms
App::update_config took 12 ms
App::update_config took 5 ms
App::update_config took 4 ms
```

PR #302, opened 2024-08-20
Zoom per tab/pane & skip writing config to fs when zooming -  considerable perf improvement when zooming

l-const:
### Before:

1.  Zooming is applied to all panes at the same time
2. Considerate amount of lag due to  the known issue when updating the config in the filesystem ( using an old SATA  Samsung SSD 840: https://www.samsung.com/us/business/support/owners/product/840-evo-series-250gb/ , maybe with an NVME with 5x write speed it is not noticable for others.)

You can observe the lag in the video: 

https://github.com/user-attachments/assets/4895faba-61c7-47f1-ae1c-2f786d3b67c0

### After

1. Zooming is applied in the active pane(entity).
2. It is really snappy.

https://github.com/user-attachments/assets/3df89562-0af8-4d94-9f75-02ae78337bb5

#### Note: I am still resetting all the panes when changing the default font size from settings adhering to the current behaviour.

https://github.com/user-attachments/assets/531441fb-caa6-4f61-9eed-29c1d56b6032

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -397,7 +397,6 @@ pub struct App {
     curr_font_weights: Vec<u16>,
     curr_font_stretch_names: Vec<String>,
     curr_font_stretches: Vec<Stretch>,
-    zoom_adj: i8,
     zoom_step_names: Vec<String>,
     zoom_steps: Vec<u16>,
     theme_names_dark: Vec<String>,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -523,6 +535,29 @@ impl App {
         self.update_config()
     }
 
+    fn update_render_active_pane_zoom(&mut self, zoom_message: Message) -> Command<Message> {
+        // skip writing config to fs when zoom in/ out
```
l-const, line 539: skipping saving config when zoomin/zoomout

jackpot51:
I've fixed #227 by efficiently saving config in https://github.com/pop-os/cosmic-term/commit/c4cb09f18394913d8e76d1a336d10241b9b1da65, now this PR needs a rebase.

Issue #297, opened 2024-08-18
Helix-editor rendered incorectly


KoreshKrasa:
helix-editor has rendering issues in cosmic-term
Comparison with alacritty (on the right). Main issue "bar" cursor rendered as block and symbol hides character.
alacritty 0.13.2
![screenshot-2024-08-18-10-56-06](https://github.com/user-attachments/assets/883823e0-a948-4d03-920e-34edef1adfea)
![screenshot-2024-08-18-10-56-28](https://github.com/user-attachments/assets/69ecd4dc-5465-4589-ab8c-9c9152232191)

A-Walrus:
Solved by #306

eduardz1:
> helix-editor has rendering issues in cosmic-term Comparison with alacritty (on the right). Main issue "bar" cursor rendered as block and symbol hides character. alacritty 0.13.2 ![screenshot-2024-08-18-10-56-06](https://github.com/user-attachments/assets/883823e0-a948-4d03-920e-34edef1adfea) ![screenshot-2024-08-18-10-56-28](https://github.com/user-attachments/assets/69ecd4dc-5465-4589-ab8c-9c9152232191)

In these screenshots it's also visible another problem that has been bugging me for a while, full screen apps such as helix are rendered with a border on the bottom and on the right. Is there an issue tracking this? Can it be fixed?

Issue #175, opened 2024-04-08
FR: support changing cursor color and set cursor shape to bar


wiiznokes:
I use `Helix` and this settings doesn't work in cosmic-term but work in other terminal.

ChocolateLoverRaj:
> Up

You don't have to comment, you can just click on the thumbs up emoji.

A-Walrus:
Solved by #306

PR #306, opened 2024-08-23
Add cursor shape support

### LINKED ISSUE #133: [BUG] cosmic-term doesn't respect DECSCUSR change cursor shape
There are escape codes to change cursor shape. These are often used in TUI apps. They currently don't work.

For example, set cursor to bar

```sh
echo -e "\e[5 q"
```
In other terminals (Gnome, kitty, xterm) running these commands makes the cursor turn into a bar.

A-Walrus:
Fixes #133, #175, and #297

[Screencast from 2024-08-23 15-03-47.webm](https://github.com/user-attachments/assets/c0d4a084-1d39-4f2e-b83b-b3c2883400e3)

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -589,28 +589,41 @@ where
         }
 
         // Draw cursor
-        terminal.with_buffer(|buffer| {
+        {
             let cursor = terminal.term.lock().renderable_content().cursor;
             let col = cursor.point.column.0;
             let line = cursor.point.line.0;
+            let color = Color::WHITE; // TODO
+            let width = terminal.size().cell_width;
+            let height = terminal.size().cell_height;
+            let top_left = view_position
+                + Vector::new((col as f32 * width).floor(), (line as f32 * height).floor());
             match cursor.shape {
                 CursorShape::Beam => {
+                    let quad = Quad {
+                        bounds: Rectangle::new(top_left, Size::new(1.0, height)),
+                        ..Default::default()
+                    };
+                    renderer.fill_quad(quad, color);
+                }
+                CursorShape::Underline => {
                     let quad = Quad {
                         bounds: Rectangle::new(
                             view_position
                                 + Vector::new(
-                                    (col as f32 * terminal.size().cell_width).floor(),
-                                    line as f32 * terminal.size().cell_height,
+                                    (col as f32 * width).floor(),
+                                    ((line + 1) as f32 * height).floor(),
                                 ),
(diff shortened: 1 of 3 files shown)


src/terminal_box.rs
```diff
@@ -587,6 +588,46 @@ where
...
+                                    (col as f32 * width).floor(),
+                                    ((line + 1) as f32 * height).floor(),
+                                ),
+                            Size::new(width, 1.0),
+                        ),
+                        ..Default::default()
+                    };
+                    renderer.fill_quad(quad, color);
+                }
+                CursorShape::HollowBlock => {} // TODO not sure when this would even be activated
```
A-Walrus, line 626: It's not one of the options in `DECSCUSR` and I couldn't find much info about it...
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h4-Functions-using-CSI-_-ordered-by-the-final-character-lparen-s-rparen:CSI-Ps-SP-q.1D81

leb-kuchen:
```rs
pub enum CursorShape {
    /// Cursor is a block like `▒`.
    #[default]
    Block,

    /// Cursor is an underscore like `_`.
    Underline,

    /// Cursor is a vertical bar `⎸`.
    Beam,

    /// Cursor is a box like `☐`.
    HollowBlock,

    /// Invisible cursor.
    Hidden,
}
```

A-Walrus: Sure, what I'm not sure about is in what case HollowBlock would be selected.

leb-kuchen: It the application overrides the default cursor.

MoSal: @A-Walrus 

You are write to note that no sequence sets this shape.

It's used by alacritty (the app) for unfocused terminals from the looks of it:
https://github.com/alacritty/alacritty/blob/5060f8eeb864e8c304fbad9588bdd882db942356/alacritty/src/display/content.rs#L57-L58

leb-kuchen: I guess eventually there will a config option for setting the cursor shape, though I am not sure whether the UX team likes hollow cursor.

jackpot51: I think it is fine to have HollowBlock not handled yet. The ability to handle underline and beam cursors is already an improvement so I will merge.

jackpot51 (approved):

snaggen:
I see the comment about the HollowBlock, that is used instead of the block cursor when the terminal doesn't have focus.

PR #313, opened 2024-08-27
Fix regression with hidden cursors

A-Walrus:
Fix regression caused by #306 (my bad ;P) where hiding cursor didn't work correctly.

Can be tested with `echo -e "\e[?25l"` which should hide the cursor.


jackpot51 (approved):

nadav26740:
👍🏾GJ

Issue #323, opened 2024-09-02
[Bug] Fails to launch if the font is missing


Elias-Graf:
```
❯ cosmic-term
[2024-09-02T20:07:41Z ERROR cosmic_term] 'Fira Mono' is not in the font list
thread 'main' panicked at src/main.rs:620:69:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Interestingly enough, the configured font should be Hasklug NF (maybe that got cleared though when upgrading the package).

Elias-Graf:
Ah, ok, not true. I misread, the error is not the panic. See panic here with backtrace:

```
❯ RUST_BACKTRACE=full cosmic-term
[2024-09-02T20:12:33Z ERROR cosmic_term] 'Fira Mono' is not in the font list
thread 'main' panicked at src/main.rs:620:69:
index out of bounds: the len is 0 but the index is 0
stack backtrace:
   0:     0x5566940b15c5 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hd736fd5964392270
   1:     0x5566940e01fb - core::fmt::write::hc6043626647b98ea
   2:     0x5566940ad00f - std::io::Write::write_fmt::h0d24b3e0473045db
   3:     0x5566940b139e - std::sys_common::backtrace::print::h45eb8174d25a1e76
   4:     0x5566940b2a89 - std::panicking::default_hook::{{closure}}::haf3f0170eb4f3b53
   5:     0x5566940b282a - std::panicking::default_hook::hb5d3b27aa9f6dcda
   6:     0x5566940b2f23 - std::panicking::rust_panic_with_hook::h6b49d59f86ee588c
   7:     0x5566940b2e04 - std::panicking::begin_panic_handler::{{closure}}::hd4c2f7ed79b82b70
   8:     0x5566940b1a89 - std::sys_common::backtrace::__rust_end_short_backtrace::h2946d6d32d7ea1ad
   9:     0x5566940b2b37 - rust_begin_unwind
  10:     0x556692d645c3 - core::panicking::panic_fmt::ha02418e5cd774672
  11:     0x556692d647d7 - core::panicking::panic_bounds_check::hc4af019f20d205d1
  12:     0x556692ec0b68 - cosmic_term::App::set_curr_font_weights_and_stretches::hef57adfd73028dc5
[...]

PR #324, opened 2024-09-02
Early exit if no monospace fonts were found

MoSal:
This also provides a helpful log error message and avoids triggering a later panic as reported in #323.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1415,6 +1415,13 @@ impl Application for App {
             });
             font_name_faces_map
         };
+
+        if font_name_faces_map.is_empty() {
+            log::error!("at least one monospace font with normal/bold weights and default stretch is required");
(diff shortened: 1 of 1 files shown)


Elias-Graf (approved):

jackpot51 (approved):

jackpot51:
I wonder if we should embed a default font to avoid this potential problem.

Elias-Graf:
That would be very nice. This seems to be the issue for every COSMIC app and libcosmic. For example, the COSMIC file manager. Ttaking a browser, nautilus, or seemlingly every other app, they seem to have a default font (or they don't completely fail if that file [/etc/fonts/fons.conf] can be read, because `fc-list` does find fonts in other places). See the liked issue for more context.

Nautilus vs COSMIC Files:

image

Issue #10, opened 2023-12-23
Open hyperlinks


jackpot51:
Add an option to open hyperlinks

simonvanderveldt:
For others that are interested, this was fixed by https://github.com/pop-os/cosmic-term/pull/379 and https://github.com/pop-os/cosmic-term/pull/395

PR #341, opened 2024-09-24
hyperlink

leb-kuchen:
This already supports OSC8 escape sequences. Though there a few [improvements](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda ) which could be made.
For the regex approach, I think it is too expensive to do a regex search in Terminal.update.

### DIFF
--- a/ansi-colors.sh
+++ b/ansi-colors.sh
@@ -90,3 +90,5 @@ do
     printf "\033[4:5mDashedUnderline\033[0m "
     printf "\n"
 done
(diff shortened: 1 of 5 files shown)


snaggen:
Would it be possible to detect if the mouse is above a line and only perform a regex on a line if it is under a mouse pointer?

simonvanderveldt:
Sounds like this would fix #10 ?

PR #365, opened 2024-10-29
i18n(ar): add Arabic translation

x9a:


i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+default-bold-font-weight = وزن الخط السميك
+use-bright-bold = جعل النص السميك ساطعًا
+
+### Splits
+splits = الانقسامات
+focus-follow-mouse = تركيز الكتابة يتبع الفأرة
+
+### Advanced
+advanced = متقدم
+show-headerbar = إظهار الترويسة
```
MoSal, line 61:
```suggestion
show-headerbar = إظهار قائمة النافذة
```
Maybe not the best suggestion, and maybe not 100% accurate. But I hate the neologism used much more.

Same for the line below.

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+show-header-description = يعرض الترويسة من خلال قائمة النقر باليمنى.
+
+# Find
+find-placeholder = اعثر...
```
MoSal, line 65:
```suggestion
find-placeholder = اعثر على...
```

Same for the next two lines.

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+
+# Menu
+
+## File
+file = ملف
+new-tab = لسان جديد
+new-window = نافذة جديدة
+profile = اللاحة
+menu-profiles = اللاحات...
+close-tab = أغلِق اللسان
```
MoSal, line 77: Same note about the words used for "profile" and "tab".

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+quit = غادِر
+
+## Edit
+edit = تعديل
```
MoSal, line 81: Why use a noun here?

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+copy = انسِخ
+paste = ألصِق
+select-all = حدِّد الكل
+find = اعثر
+clear-scrollback = امحِ التمرير الخلفي
```
MoSal, line 86: Is this known terminology?

Otherwise, maybe something like

```suggestion
clear-scrollback = امحِ الصفحات السابقة
```

or even if one goes the literal literal route

```suggestion
clear-scrollback = امحِ اللفائف السابقة
```

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+
+## View
+view = عرض
+zoom-in = نص أكبر
```
MoSal, line 90: Use verbs. Same for the next two.

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+zoom-reset = حجم النص المبدئي
+zoom-out = نص أصغر
+next-tab = اللسان التالي
+previous-tab = اللسان السابق
+split-horizontal = اقسم أفقيًا
+split-vertical = اقسِم عموديًا
+pane-toggle-maximize = بدِّل التكبير
```
MoSal, line 97: A better word for maximize if exists.

i18n/ar/cosmic_term.ftl
```diff
@@ -0,0 +1,100 @@
...
+
+## Profiles
+profiles = اللاحات
+name = الاسم
+command-line = سطر الأوامر
+tab-title = عنوان اللسان
+tab-title-description = يتجاوز عنوان اللسان المبدئي
+add-profile = أضِف لاحة
+new-profile = لاحة جديدة
+make-default = اجعله المبدئي
```
MoSal, line 25: I thought we used "قياسي" for "default", or was it for "standard"?

(I don't like either anyway.)

jackpot51 (approved):

MoSal (changes requested):
Merged already, I see.

Anyway, here is somewhat picky suggestions from someone who doesn't use translated interfaces.

I used the "how easy can I back-translate this", and "how easy can I guess what is this referring to", which is probably not ideal where common translated terminology (unknown to me) is used. So feel free to ignore most of them.

There are a few changes that should be done anyway, if you can do this in a followup PR @x9a.

PR #408, opened 2025-01-08
Fix broken numlock behavior

Nashenas88:
libcosmic broke backwards compatibility in pop-os/libcosmic@0491c4b
Add comparison with modified_key to ensure only actual named keys are processed as such.

This is related to pop-os/libcosmic#702


Nashenas88:
It's not clear to me whether this is the right approach. I tried looking into fixing the behavior in libcosmic, but it seems more like a behavior change and break in contract than a bug in the lib.

PR #426, opened 2025-01-30
Add very basic custom keybinding support

Simply56:
The shortcuts are read from a yaml file located at `~/.config/cosmic-term/cosmic-term.yaml`.
Keys that would have to be accessed using `Key::Named()` are not yet supported.

I am new to contributing so I am open to suggestions :D

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1278,6 +1278,7 @@ dependencies = [
  "env_logger 0.10.2",
  "fork",
  "hex_color",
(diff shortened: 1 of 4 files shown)


mmstick:
We already have a config file mechanism in [cosmic-config](https://pop-os.github.io/libcosmic/cosmic/cosmic_config/index.html), which all COSMIC apps use for their configuration data.

PR #442, opened 2025-02-27
add search flags ( whole words and match case)

YkdWaWEzVmphR1Z1:
Screenshot_2025-02-28_00-10-49


jackpot51:
Thanks for this! Yeah, I'd like regex to be optional as well.

leb-kuchen:
image

I applied the style and wording of cosmic-edit, but now with 2 rows you notice how the padding is a bit off.

PR #454, opened 2025-03-17
Add --version and --help with clap_lex

LinuxBoy-96:
This is a better version of my previous attempt.
This ensures that cosmic-term can provide its version and some information via the terminal.
This is essential for when multiple distributions start using COSMIC.

Clap has the benefit of managing the version automatically by reading Cargo.toml.

Note that my objective is to implement it in most COSMIC apps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ shlex = "1"
 tokio = { version = "1", features = ["sync"] }
 # CLI arguments
 lexopt = "0.3.0"
-vergen = "9"
 # Internationalization
 i18n-embed = { version = "0.15", features = [
     "fluent-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ serde = { version = "1", features = ["serde_derive"] }
 shlex = "1"
 tokio = { version = "1", features = ["sync"] }
 # CLI arguments
-clap = "4"
+lexopt = "0.3.0"
 # Internationalization
 i18n-embed = { version = "0.15", features = [
     "fluent-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ serde = { version = "1", features = ["serde_derive"] }
 shlex = "1"
 tokio = { version = "1", features = ["sync"] }
 # CLI arguments
-lexopt = "0.3.0"
+clap_lex = "0.7"
 # Internationalization
 i18n-embed = { version = "0.15", features = [
     "fluent-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ ron = "0.8"
 serde = { version = "1", features = ["serde_derive"] }
 shlex = "1"
 tokio = { version = "1", features = ["sync"] }
+# CLI arguments
+clap = "4"
 # Internationalization
 i18n-embed = { version = "0.15", features = [
     "fluent-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ shlex = "1"
 tokio = { version = "1", features = ["sync"] }
 # CLI arguments
 lexopt = "0.3.0"
+vergen = "9"
 # Internationalization
 i18n-embed = { version = "0.15", features = [
     "fluent-system",
--- a/src/main.rs
+++ b/src/main.rs
@@ -108,10 +108,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 return Ok(());
             }
             Some("--no-daemon") => {
-                // Only disable daemonization under Redox
-                if cfg!(target_os = "redox") {
-                    daemonize = false;
-                }
+                daemonize = false;
             }
             Some("-e") | Some("--command") => {
                 // Handle the '--command' or '-e' flag
--- a/src/main.rs
+++ b/src/main.rs
@@ -68,6 +68,8 @@ mod terminal_theme;
 
 mod dnd;
 
+const APP_VERSION: &str = env!("CARGO_PKG_VERSION");
+
 lazy_static::lazy_static! {
     static ref ICON_CACHE: Mutex<IconCache> = Mutex::new(IconCache::new());
 }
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -80,6 +82,19 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
 /// Runs application with these settings
 #[rustfmt::skip]
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Add CLI arguments managements with `clap`
+    let matches = clap::Command::new("cosmic-term")
+        .version(env!("CARGO_PKG_VERSION"))
```
jackpot51, line 87: The version is Cargo.toml is not being actively used at this point. I'd recommend to look at the `vergen` information that is shown on the about page.

LinuxBoy-96: @jackpot51 well it's not a big difference but I this is how I view it for now. 
And honestly like it better if the output is simplier.

src/main.rs
```diff
@@ -69,10 +69,27 @@ mod terminal_theme;
...
+use lexopt::{Parser, Arg};
+const APP_VERSION: &str = env!("CARGO_PKG_VERSION");
+
 lazy_static::lazy_static! {
     static ref ICON_CACHE: Mutex<IconCache> = Mutex::new(IconCache::new());
 }
 
+fn print_help() {
+    println!(
+        r#"
```
mmstick, line 81: The binary name should be set on the same line as `r#"` or else there will be an extra newline on top.

I also think there's no reason not to show the current version on the help page alongside the name.

Comparing with `rg -h` for example, I'd expect to see

```
cosmic-term {{APP_VERSION}} (git commit {{VERGEN_GIT_SHA}} on ({{VERGEN_GIT_COMMIT_DATE}})
System76 <info@system76.com>

{{description}}

OPTIONS:
    -h, --help      Show help output.
    -v, --version  Print cosmic-term's version.
```

src/main.rs
```diff
@@ -81,6 +98,23 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
 /// Runs application with these settings
 #[rustfmt::skip]
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut parser = Parser::from_env();
+
+    // Parse the arguments
+    while let Some(arg) = parser.next()? {
+        match arg {
+            Arg::Long("help") => {
```
mmstick, line 106:
```rs
Arg::Short('h') | Arg::Long("help")
```

src/main.rs
```diff
@@ -81,6 +98,23 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
...
+                print_help();
+                return Ok(());
+            }
+            Arg::Long("version") => {
```
mmstick, line 110:
```rs
Arg::Short('v') | Arg::Long("version")
```

src/main.rs
```diff
@@ -81,6 +98,23 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
...
+                println!("cosmic-term {}", APP_VERSION);
```
mmstick, line 111: Needs the vergen git info to be useful. Helps to know what git revision is being used when a bug report is given.

LinuxBoy-96: I do not really like how messy it look with vergen but it's there now...

Cargo.toml
```diff
@@ -25,6 +25,9 @@ ron = "0.8"
 serde = { version = "1", features = ["serde_derive"] }
 shlex = "1"
 tokio = { version = "1", features = ["sync"] }
+# CLI arguments
+lexopt = "0.3.0"
+vergen = "9"
```
mmstick, line 30: Not needed. This is already applied in the build.rs: https://github.com/pop-os/cosmic-term/blob/master/build.rs

src/main.rs
LinuxBoy-96: Newline fixed, but no --help  never has the version included.

mmstick: Not true.

```
$ rg -h
ripgrep 14.1.0
Andrew Gallant <jamslam@gmail.com>

ripgrep (rg) recursively searches the current directory for lines matching
a regex pattern. By default, ripgrep will respect gitignore rules and
automatically skip hidden files/directories and binary files.

Use -h for short descriptions and --help for more details.

Project home page: https://github.com/BurntSushi/ripgrep

USAGE:
  rg [OPTIONS] PATTERN [PATH ...]

POSITIONAL ARGUMENTS:
  <PATTERN>   A regular expression used for searching.
  <PATH>...   A file or directory to search.
```

LinuxBoy-96:
Well it seem fine now.

LinuxBoy-96:
This feature now has the same structure as in cosmic-greeter, which has been merged and cosmic-player.

jackpot51:
What happened to the parsing of --no-daemon?

LinuxBoy-96:
> What happened to the parsing of --no-daemon?

it's still there after my code, between line 111 and 143.

However @mmstick did a great job with cosmic-greeter, puting my code in argparse.rs. So that's an options if you don't want that code there.

jackpot51:
I do think the argument parsing must be unified and the help text must mention all possible flags and arguments.

LinuxBoy-96:
> I do think the argument parsing must be unified and the help text must mention all possible flags and arguments.

So I did a thing, hope it's not that bad... I have no objection into having a argparse.rs files from that point.

LinuxBoy-96:
@jackpot51 My stripped binary size is 36120 KB, 16 BB smaller than the official. So it seem like a clean and efficient way to do it.

If you like it, maybe using vergen and clap_lex should be standardized across all COSMIC apps.

src/main.rs
```diff
@@ -80,33 +84,59 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
...
-                _ => {
-                    //TODO: should this throw an error?
-                    log::warn!("ignored argument {:?}", arg);
-                }
             }
-        } else if shell_program_opt.is_none() {
-            shell_program_opt = Some(arg);
+            Some("-e") | Some("--command") => {
+                // Handle the '--command' or '-e' flag
+                continue;
```
jackpot51, line 118: This should break instead of continue

LinuxBoy-96: @jackpot51 It's fixed

src/main.rs
```diff
@@ -79,31 +83,60 @@ pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
...
+            Some("--version") | Some("-V") => {
+                println!(
+                    "cosmic-term {} (git commit {})",
+                    env!("CARGO_PKG_VERSION"),
+                    env!("VERGEN_GIT_SHA")
+                );
+                return Ok(());
+            }
+            Some("--no-daemon") => {
+                // Only disable daemonization under Redox
```
jackpot51, line 111: Why is this gated to only Redox?

LinuxBoy-96: > Why is this gated to only Redox?

Fixed now, I got confuse at some point because theire is no deamonization on Redox.

PR #471, opened 2025-04-24
feat(color-schemes): add Seashells color-scheme

captain-maramo:
adding Seashells color-scheme to make it availabe


captain-maramo:
here is a little example, what it looks like
Screenshot_2025-04-24_11-01-13

jackpot51 (approved):

PR #478, opened 2025-05-04
Add support for XF86Copy and XF86Paste keybindings

markstos:
Keyboards with custom firmware that support dedicated "Copy" and "Paste" keycodes are more common now.

These keycodes solve a pain point with terminals: that you have to use Control+Shift+C/V to paste into the terminal, but Control+C/V everywhere else.

The dedicated Copy/Paste keycodes solve this problem by allowing you to use the same key combo in all apps that support it.

This patch supports the additional bindings by default while leaving the existing keybindings untouched.


jackpot51 (approved):

jackpot51:
Here did you get these key names from? This change does not appear to compile.

git-f0x:
Browsing the iced source, maybe just `Copy` and `Paste` will work?

PR #481, opened 2025-05-06
feat: bind copy/paste keycodes to copy/paste actions.

markstos:
This PR replaces #478, which was not tested properly.

To add the keycodes bindings without modifiers, it was necessary to update the bind()
macro to support that. I followed the precedent from cosmic-files to do that. 

## Testing

 - Compiled cosmic-term, ran it, tested pasting with the Paste keycode.


markstos:
@jackpot51 Thanks for the feedback. I found the precedent from cosmic-files, implemented and tested it. It's a better solution!

jackpot51 (approved):

PR #485, opened 2025-05-09
i18n(nl) update Dutch translation

feikedonia:


i18n/nl/cosmic_term.ftl
```diff
@@ -25,7 +25,7 @@ new-profile = Nieuw profiel
 make-default = Als standaard instellen
 working-directory = Werkmap
 hold = Behouden
-remain-open = Venster geopend houden nadat het subproces voltooid is.
+remain-open = Dit venster openhouden nadat het kind-proces eindigt.
```
feikedonia, line 28: (EN): Remain open after child process exits.
what remains open?
I have translated it as
(NL-lit.): This window remains open after the child process finishes.

jackpot51 (approved):

PR #499, opened 2025-06-03
Use `LazyLock` instead of `lazy_static!`

Doods0:
As-salamu alaykum

While checking out Clippy's pedantic lints on this repo, I noticed [this](https://rust-lang.github.io/rust-clippy/master/index.html#non_std_lazy_statics).

As linked above, there are two obvious benefits to this:
>Reduces the need for an extra dependency.
>Enforce convention of using standard library types when possible.


jackpot51 (approved):

jackpot51:
Looks good but needs a rebase

Issue #504, opened 2025-06-15
Support F11 Key to enter fullscreen mode


atiq-cs:
Feature request:

Like many other popular terminals, can we support enabling fullscreen mode when key F11 is pressed and toggles out of it if it is pressed again?

PR #505, opened 2025-06-19
Add F11 support

liberodark:
Fix : https://github.com/pop-os/cosmic-term/issues/504


jackpot51:
Perhaps Ctrl+Shift+F with an added menu item, or Shift+F11?

liberodark:
> Perhaps Ctrl+Shift+F with an added menu item, or Shift+F11?

Hi,

I understand your point of view about that.
But also im see other most terminal support F11 switch in Fullscreen.
Afterwards, I'll follow your preferences.
But I think that not losing the user and having the same shortcuts also improves the customer experience.
PS: A quick note that has nothing to do with it, but congratulations for posix-regex

Best Regards

jackpot51:
alacritty and xterm appear to pass F11 through, gnome-terminal fullscreens instead. I believe `screen` uses F11 as a key binding, but looking through, I don't think it has affected many users for gnome-terminal to not allow it. I can accept this as-is and see if there are complaints.

PR #513, opened 2025-06-30
Update Cargo.toml to strip the release binary

LinuxBoy-96:
Strip the binary for the release build.


mmstick:
This is already being done by packaging tools on packaging. Sometimes even generating -dbg packages from the stripped symbols, which this would remove.

PR #574, opened 2025-10-02
Reconsider the quick password access support

snaggen:
Hi

I have been maintaining this "password manger" feature in my own branch, but it would be good for things like i18n and general discoverability to get it merged. 

Last time it was said that a password management feature is not in line with the product and that users can use the regular password manager (like seahorse) instead. To this, I would like to point out that it is not really much of a manager, that is not the feature. The feature is to provide quick and easy access to insert passwords for ssh, vpn and similar. The passwords are already stored in your keyring, so this just provides quick access to them. Much like the auto-fill feature in iOS and android. 

So, the workflow for the user to start a cmdline based vpn will be: 
1. start vpn
2. Ctrl+Alt+p 
3. Look at a short list of passwords and click on it.

Instead of:
1. start vpn
2. Open Seahorse
4. Search through a large list of passwords and open the correct one
5. Copy the password
6. Switch to the terminal again and paste

If you work with many VPNs or other tasks tasks that requires a lot of passwords, this feature will make your life a lot easier. 

The downside to include this is mainly more code to maintain. But since I have been maintaining it in it's own branch, I have structured it to be easy to maintain so that should not be a very big problem.


jackpot51:
Thanks, it looks better but for portability, it should be behind a feature gate.

snaggen:
I will add a feature gate

jackpot51:
You could consider making a Message enum in the password_manager.rs file and then embedding it in a Password message at the higher level. That would make it more modular.

jackpot51:
Then, something like this would be all that is needed in the update function, and any added messages would not all need feature gates.

```
Message::PasswordManager(password_manager_message) => {
    self.password_manager.update(password_manager_message);
}
```

snaggen:
Will move more stuff in to password manager...

src/password_manager.rs
```diff
@@ -0,0 +1,271 @@
...
+                    .width(Length::Fixed(320.0))
+                    .push(
+                        widget::button::text(label.clone())
+                            .width(Length::Fixed(280.0))
+                            .on_press(Message::PasswordManager(
+                                PasswordManagerMessage::PasswordFetch(label.clone()),
+                            )),
+                    )
+                    .push(
+                        widget::button::text("-")
```
jackpot51, line 130: Please use the icon button used for deleting profiles.

jackpot51:
The delete button is the last thing I saw, but UX may have other changes.

snaggen:
Thanks for the mockup! I wil try to implement this during the weekend

snaggen:
I also changed the title from "Manage Passwords" to just "Passwords" since the main use of the page is to access and paste passwords, the management is still possible but that is not what it is mainly used for.

jackpot51:
Thanks, I will review on Monday

PR #579, opened 2025-10-03
chore: update libcosmic

wash2:


git-f0x:
It would maybe be good to add the `license_url` for about, like in Files.

jackpot51 (approved):

PR #609, opened 2025-10-29
terminal: auto-scroll drag selection

stephanbuys:
This adds auto-scroll while making a selection by dragging beyond the visible viewport. When the pointer is above/below the view during a drag-selection, the view scrolls in that direction until the pointer re-enters the viewport or the selection ends. This matches expected behaviour from other terminal emulators and editors.

Details
- Scrolls at a moderated rate to avoid overshoot.
- Active only during drag-selection; no impact on normal scrolling.

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -17,6 +17,7 @@ use cosmic::{
         event::{Event, Status},
         keyboard::{Event as KeyEvent, Key, Modifiers},
         mouse::{self, Button, Event as MouseEvent, ScrollDelta},
+        window::RedrawRequest,
     },
     iced_core::{
         Border, Shell,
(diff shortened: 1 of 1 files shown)


wash2 (approved):

jacobgkau (commented):
When testing, I'm noticing this:

- If I click and drag past the edge of the window, the view only scrolls while the mouse cursor's moving, not while it's still. This feels fine as expected behavior.
- However, when I release the mouse button with the cursor still outside the window, the selection jumps to include (or not include, if I'm de-selecting) a few extra lines.

For comparison, GNOME Terminal doesn't have that jump when releasing the mouse button-- it stops selecting immediately when the mouse button's released (although it also continues selecting more for the entire time the cursor's held outside the window, as long as the mouse button's held down, without the cursor needing to be moved; it only gets faster when moving the cursor).

One way or another, it seems like the jump on release is a bug and should be eliminated.

jacobgkau (commented):
I'm still seeing the jump on mouse release (while the mouse is above or below the window) on c0d4131.

To be clear, the steps to reproduce are:

1. `sudo dmesg` to generate long output
2. Click and start dragging from inside the window to above the window
3. Stop moving the mouse cursor & wait for the text to stop scrolling
4. Release the mouse button; this is when the jump occurs

jacobgkau:
Actually, if I'm careful, it's often not happening now. The times it's happening might just be me moving the cursor slightly as I let go of it.

wash2 (approved):

stephanbuys:
That's good to hear, there's some things that could be tried, maybe this is a ok start and see what the user feedback is more generally?
Otherwise if there's a terminal that definitely works in with the trackpad then maybe I can look at how they do it?

stephanbuys:
This latest attempt takes how Konsole and Gnome's terminal handle the auto scroll behaviour into account. Hope this settles it. I unfortunately don't have a trackpad, Konsole or gnome - so this is hard to reproduce and fully verify for me, but initial testing seems fine.

jacobgkau (approved):
Alright, a404840 continues scrolling (at a more manageable speed) as long as the cursor's out of the window, so that eliminates the real use-case for letting go of the cursor outside of the window for precision operations. I also don't see a noticeable jump happening anymore, so that's great.

This seems to be working well with both a touchpad and a mouse. Both normal select and word select still work, as do the selection keyboard shortcuts. Also seems to work as expected when new output's appearing while I'm selecting. Not seeing any regressions.

Thank you for implementing this!

PR #619, opened 2025-11-06
Update dependencies

git-f0x:
Gets the new context drawer design, and switches `alacritty_terminal` to the crate version, rather than an old git commit. The only breaking change since then is `hold` being renamed to `drain_on_exit`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,8 +139,9 @@ dependencies = [
 
 [[package]]
 name = "alacritty_terminal"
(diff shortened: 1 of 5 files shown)


jackpot51 (approved):

git-f0x:
Yeah, I read the changelog and saw:
>Replaced `Options::hold` with `Options::drain_on_exit`

so it seems I kind of misunderstood what that meant (and seemingly didn't read the commit description of the commit I linked 🫠).

PR #662, opened 2025-12-24
feat: add `--working-directory` arg

### LINKED ISSUE #636: cosmic-term doesn't support a working directory as parameter
It is not possible to use the Caja action "Open in terminal" with cosmic term because it doesn't support the working directory as parameter.
For instance you can try
```bash
cosmic-term ~/Music
```
and see that the new terminal is not opened in the expected directory

adil192:
Adds a `--working-directory` argument.

I want this feature so I can use the "Open in terminal" button in GitHub Desktop.

I chose `--working-directory` to match `gnome-terminal`'s equivalent argument, and added a `-w` shorthand. Feel free to change these.

Also I'm new to Rust so please forgive any rookie mistakes (thank you to ada4a for your feedback) 

Resolves #636 

***

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1523,14 +1523,13 @@ impl App {
                     Some(colors) => {
                         let current_pane = self.pane_model.focused();
                         if let Some(tab_model) = self.pane_model.active_mut() {
-                            let mut options = self.startup_options.take().unwrap_or_default();
-                            let mut tab_title_override = None;
-                            if let Some(profile) = profile_id_opt
+                            let (options, tab_title_override) = if let Some(profile) = profile_id_opt
                                 .and_then(|profile_id| self.config.profiles.get(&profile_id))
                             {
                                 // Merge profile and startup options, preferring startup options
-                                options = tty::Options {
-                                    shell: options.shell.or_else(|| {
+                                let startup_options = self.startup_options.take().unwrap_or_default();
+                                let options = tty::Options {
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau (changes requested):
Choosing a profile from the menu still works, but setting a profile as default no longer applies that profile upon launch on this branch. It works on the currently released version of COSMIC Term.

To make sure I didn't break this when I rebased and resolved the conflict today, I checked out [42ddc92](https://github.com/pop-os/cosmic-term/pull/662/commits/42ddc92370284fc7dce9fa44083b5862bf0a5161) and built that, and the profile set as default still didn't apply on launch. After installing the released version, the profile's applied on launch again.

adil192:
> setting a profile as default no longer applies that profile upon launch on this branch. 

Thanks for finding that bug. I've fixed it by merging the profile options and startup options.

For working_directory and shell, I've used (pseudo code) `startup_options.value.or_else(|| profile.value)` to prefer startup_options's value if provided.

However, drain_on_exit is a bool and not an Option, so I've just used an OR statement. Is that the desired behaviour? I don't really know what drain_on_exit is used for.
```rust
drain_on_exit: options.drain_on_exit || profile.drain_on_exit,
```

jacobgkau (approved):
The broken "hold" option aside, the flag/profile merging is working as expected now (with the `--working-directory` and `--command` options). Also still works without a default profile set.

PR #673, opened 2025-12-29
Drop down mode

timbz:
This PR adds a drop-down mode, similar to terminals like Guake or Tilda. This feature is substantial to my workflow and I thought others might find it useful as well.

To use it, create a custom shortcut in COSMIC Settings with the command `cosmic-term --drop-down` . 
The first invocation launches the drop-down terminal, subsequent calls toggle it show/hide. Only one instance runs at a time.

I'm still learning Rust, so please be gentle with feedback. I'd love to hear if this is something you'd be interested in merging, or if you'd prefer a different approach for this functionality. Happy to iterate on any suggestions for improvement!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1417,7 +1417,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic.git#3b8ad45950f5d23c8550e18e628f6e70b7089d89"
+source = "git+https://github.com/pop-os/libcosmic.git#dd3610b8ae4f1bcf2e2299e82f908913d1a4a57d"
 dependencies = [
  "atomicwrites",
  "cosmic-config-derive",
@@ -1438,7 +1438,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config-derive"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic.git#3b8ad45950f5d23c8550e18e628f6e70b7089d89"
+source = "git+https://github.com/pop-os/libcosmic.git#dd3610b8ae4f1bcf2e2299e82f908913d1a4a57d"
 dependencies = [
  "quote",
  "syn 2.0.109",
@@ -1447,7 +1447,7 @@ dependencies = [
 [[package]]
 name = "cosmic-files"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-files.git#e779fc3dac1dd1de2b5c8aa54c6bd5f872169635"
+source = "git+https://github.com/pop-os/cosmic-files.git#d9b6404f1b68e8cef5d879c98ddf96c9e7207196"
 dependencies = [
  "anyhow",
  "chrono",
@@ -1541,6 +1541,7 @@ dependencies = [
  "clap_lex",
  "cosmic-files",
  "cosmic-text",
+ "dirs 5.0.1",
  "fork",
  "hex_color",
  "i18n-embed",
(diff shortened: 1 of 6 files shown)


dickwall:
Hang in there @timbz - the lack of a decent drop down terminal is one of the things preventing me switching my main development machine to cosmic, and I'm hopeful this will be accepted or at least spur development in this missing feature. It's discouraging to see that there has been no acknowledgement from the devs though :-(

rnbwdsh:
i rebased and refactored this, i'd be happy for any feedback, especially on the comments I put there

asilenkov:
> i rebased and refactored this, i'd be happy for any feedback, especially on the comments I put there

Hello! I've tried this on Fedora 44, works for me, thanks for your efforts.
There is one small glitch,  

Bug in find_existing_dropdown_pid – single-instance toggle always fails
src/dropdown.rs:23 — the ? operator causes the function to return None on the first non-numeric /proc entry (e.g. cpu, self, net), so every F12 press spawns a new cosmic-term --drop-down process instead of sending SIGUSR1 to the existing one.
Fix: Replace ? with continue so non-PID directories are silently skipped.
// Before (line 23):
let pid = entry.file_name().to_string_lossy().parse::<u32>().ok()?;

// After:
let pid = match entry.file_name().to_string_lossy().parse::<u32>() {
    Ok(pid) => pid,
    Err(_) => continue,
};
Also applies to the cmdline read on the next line — ok()? should be continue, not ?, for the same reason.

asilenkov:
more on this 
- top bar with File, Edit, View is not working at all, Layer surface can't create popup submenus. 
- fonts filter are a little bit not optimal. Filter drops all the fonts which don't have Bold variant. Better to let it make synthetic bold befor filtering. 
Everything else is quite good, thanks for this.

rnbwdsh:
> Hello! I've tried this on Fedora 44, works for me, thanks for your efforts.....

I'm not sure i understand the feedback regarding the `find_existing_dropdown_pid` as that is a function name I use in my branch / pull request: https://github.com/pop-os/cosmic-term/pull/881 and is not used in this branch.

But my `find_existing_dropdown_pid` doesn't use any ? and isn't in a loop on dropdown.rs:23 that could be continued, line 23 is the `Err(_) => return None,` part of the read_dir so the error handling if you don't have a readable /proc.

Is it possible you are on an old version? I have force push squashed the branch as there was some messy stuff inbetween.

> fonts filter are a little bit not optimal

I don't understand what you mean here, neither my nor timbz code interacts with fonts at all? (Even though a very strong font filtering currently seems to exist, i can confirm that but that's not related to this branch?

> top bar with File, Edit, View is not working at all, Layer surface can't create popup submenus.

I'll quickly work that one out, i had that already working in the process but I think I broke it in some refactoring. For some reason, the mouse over effect of these buttons already works (see screenshot in my mr), they just don't seem to properly react any more

rnbwdsh:
i managed to fix the menu bar in a diff as minimal as possible (i think) as "Layer surface can't create popup submenus" is not fully correct for cosmic.

Sadly, I can't screenshot it (with cosmic-screenshot, at least) because the screenshot tool overlay interfers with the surface menu overlay.

https://github.com/pop-os/cosmic-term/pull/881/changes/c748bf3721225e051ef57fbca47c1d05f2c3efe1

Edit: I also rebased on the current master and it'd be ready to review/merge.

asilenkov:
> I'm not sure i understand the feedback regarding the `find_existing_dropdown_pid`

Sorry for a confusion, probably I messed with git a bit, that's probably another PR not yours, sorry. 
Font filter — pre-existing in master, not part of any drop-down PR, just a small fix to handle fonts better. 
And menubar is probably libcosmic glitch

rnbwdsh:
no the menu bar was a real problem, because a regular responsive_menu_bar has the problems you described, but a menu::bar is fine, so that was a real issue / regression

PR #674, opened 2025-12-30
make hotkeys configurable

### LINKED ISSUE #197: Option to change keyboard shortcuts
I'm guessing that there is already an issue about this in a different repo but idk. I would like the ability to re-bind paste to Ctrl+V.

nick1udwig:
## Problem

```
//TODO: load from config
```

The hotkeys are hardcoded! This means that, e.g., I cannot use my `F11` key for anything other than maximizing my terminal.

## Solution

Add a hotkey file inspired by cosmic-settings. Add a menu to the settings GUI for configuring away from defaults.


jackpot51 (commented):
A new config id is not needed, the shortcuts could be in the original id with shortcuts_ prefixed

src/shortcuts.rs
```diff
@@ -0,0 +1,468 @@
...
+            };
+            (Some(config_handler), config)
+        }
+        Err(err) => {
+            log::error!("failed to create shortcuts config handler: {}", err);
+            (None, ShortcutsConfig::default())
+        }
+    }
+}
+
```
jackpot51: These will need to be translateable. There are already translations for most.

nick1udwig:
> A new config id is not needed, the shortcuts could be in the original id with shortcuts_ prefixed

I tried to address this in d5f1ac4

See other comments in threads above; pinging for another round of review

jackpot51 (dismissed):

leviport (approved):
Here is a clip of the scroll jumping behavior when the confirm prompt appears:

https://github.com/user-attachments/assets/f71982de-e754-41ba-801f-239b04332d7e

It's a bit jarring, but I don't think it needs to block merging. No scroll jump happens when there is no hotkey conflict. @maria-komarova what are your thoughts on this? I have it pushed to `popdev:testing-cosmic-term-pr674` if you want to test it yourself.

If we're alright with the scroll jump when the confirmation dialog appears for now, I can open a tracking issue and it can be worked on later.

jackpot51 (approved):

gene1wood:
The issue around "Ctrl + PAGEDOWN" and "Ctrl + PAGEUP" was opened in #727

PR #678, opened 2025-12-31
make new terminal open to same dir, not home

nick1udwig:
## Problem

Creating a new terminal, e.g. with `Ctrl + Shift + N`, opens to home dir.

popos 22.04 had a more intuitive behavior of opening to the cwd of the existing terminal. This is desirable because it is trivial to change to home dir (`cd`) but potentially a lot of typing to change dir to some deeply nested directory

## Solution

Grab existing terminal cwd and set new terminal to that

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -603,6 +603,14 @@ impl App {
         Task::none()
     }
 
+    fn active_terminal_cwd(&self) -> Option<std::path::PathBuf> {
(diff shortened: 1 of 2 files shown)


eskimor (approved):
Looks good to me. No Windows support, but falls back to old behavior - so not any worse.

mati865:
Obsoleted by https://github.com/pop-os/cosmic-term/pull/825

Issue #265, opened 2024-07-31
Scrollback Buffer should be user configurable


bitFiedler:
I miss the option to set the scrollback buffer.
For me 100000 lines are too less. 

alacritty supports it, but it limited to 100000, would be nice to can use more than that.

leb-kuchen:
https://github.com/pop-os/cosmic-term/blob/72391e782da30038d806413660662f775d384cc1/src/terminal.rs#L219-L307
`Config.scrolling_history` sets the limit.
If you wanted to make a PR, you'd  have to add the scrolling_history field to the Profile struct in config.rs.
In main.rs,  App.profiles, you'd have to add a slider.  Gnome terminal has a slider with a step of 1000. 
Here is a reference PR https://github.com/pop-os/cosmic-term/pull/181.
Note, I cannot guarantee you it gets merged. You should probably talk with @jackpot51  first.

oscardssmith:
Bumping the request on this. Losing error messages to debug printing pushing the error above the scrollback is really annoying.

KyunLFA:
Right now I can't even read a long dmesg log... Bumping this as well!

PR #684, opened 2026-01-03
Make scrollback history length configurable

smessmer:
## Summary

Add per-profile configurable scrollback history setting, allowing users to control how many lines of terminal output are retained in the scroll buffer.

Addresses #265.

Written with AI help.

screenshot

### `src/config.rs`
- Add `DEFAULT_SCROLLBACK_HISTORY` constant (10,000 lines)
- Add `scrollback_history: u32` field to `Profile` struct with `#[serde(default)]` for backward compatibility

### `src/main.rs`
- Add `SCROLLBACK_HISTORY_VALUES` and `SCROLLBACK_HISTORY_NAMES` statics for dropdown options
- Add `Message::ProfileScrollbackHistory` variant and handler
- Add dropdown UI in profile settings
- Clone base `term_config` and override `scrolling_history` from profile when creating terminals

### `i18n/en/cosmic_term.ftl`
- Add `scrollback-history` and `scrollback-history-description` translations

## How It Works

1. Each profile stores its own `scrollback_history` value (default: 10,000)
2. Users select from predefined values via dropdown: 1,000 / 5,000 / 10,000 / 25,000 / 50,000 / 100,000
3. When creating a terminal, the base `term_config` is cloned and `scrolling_history` is overridden with the profile's value
4. Existing configs without the field use the default (serde default)

## Limitations

- Changes only apply to newly created terminals (alacritty_terminal limitation)
- Maximum 100,000 lines (alacritty's documented limit)

## Test Plan

- [x] Create a new profile and verify scrollback dropdown appears
- [x] Change scrollback value and open a new terminal with that profile
- [x] Generate output exceeding the limit and verify older lines are discarded
- [x] Verify existing configs load correctly with default value

## PR Template
If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
 - [x]  understand these changes in full and will be able to respond to review comments.
- [x]  I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/i18n/en/cosmic_term.ftl
+++ b/i18n/en/cosmic_term.ftl
@@ -27,6 +27,8 @@ make-default = Make default
 working-directory = Working directory
 hold = Hold
 remain-open = Remain open after child process exits.
+scrollback-history = Scrollback history
+scrollback-history-description = Maximum lines kept in scroll buffer (0-100,000). Changes apply to new terminals.
 
 ## Settings
 settings = Settings
--- a/src/config.rs
+++ b/src/config.rs
@@ -16,6 +16,7 @@ use crate::{fl, localize::LANGUAGE_SORTER};
 pub const CONFIG_VERSION: u64 = 1;
 pub const COSMIC_THEME_DARK: &str = "COSMIC Dark";
 pub const COSMIC_THEME_LIGHT: &str = "COSMIC Light";
+pub const DEFAULT_SCROLLBACK_HISTORY: u32 = 10_000;
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum AppTheme {
@@ -200,6 +201,12 @@ pub struct Profile {
     pub working_directory: String,
     #[serde(default)]
     pub drain_on_exit: bool,
+    #[serde(default = "default_scrollback_history")]
+    pub scrollback_history: u32,
+}
+
+fn default_scrollback_history() -> u32 {
+    DEFAULT_SCROLLBACK_HISTORY
 }
 
 impl Default for Profile {
@@ -212,6 +219,7 @@ impl Default for Profile {
             tab_title: String::new(),
             working_directory: String::new(),
             drain_on_exit: false,
+            scrollback_history: DEFAULT_SCROLLBACK_HISTORY,
         }
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,8 +40,8 @@ use std::{
 use tokio::sync::mpsc;
 
 use config::{
-    AppTheme, CONFIG_VERSION, ColorScheme, ColorSchemeId, ColorSchemeKind, Config, Profile,
-    ProfileId,
+    AppTheme, CONFIG_VERSION, ColorScheme, ColorSchemeId, ColorSchemeKind, Config,
+    DEFAULT_SCROLLBACK_HISTORY, Profile, ProfileId,
 };
 mod config;
 mod mouse_reporter;
@@ -75,6 +75,9 @@ mod dnd;
 use clap_lex::RawArgs;
 
 static ICON_CACHE: LazyLock<Mutex<IconCache>> = LazyLock::new(|| Mutex::new(IconCache::new()));
+
+static SCROLLBACK_HISTORY_VALUES: [u32; 6] = [1_000, 5_000, 10_000, 25_000, 50_000, 100_000];
+static SCROLLBACK_HISTORY_NAMES: [&str; 6] = ["1000", "5000", "10000", "25000", "50000", "100000"];
 
 pub fn icon_cache_get(name: &'static str, size: u16) -> widget::icon::Icon {
     let mut icon_cache = ICON_CACHE.lock().unwrap();
@@ -388,6 +391,7 @@ pub enum Message {
     ProfileNew,
     ProfileOpen(ProfileId),
     ProfileRemove(ProfileId),
+    ProfileScrollbackHistory(ProfileId, usize),
     ProfileSyntaxTheme(ProfileId, ColorSchemeKind, usize),
     ProfileTabTitle(ProfileId, String),
     Surface(surface::Action),
@@ -920,6 +924,9 @@ impl App {
                         .theme_names_light
                         .iter()
                         .position(|theme_name| theme_name == &profile.syntax_theme_light);
+                    let scrollback_selected = SCROLLBACK_HISTORY_VALUES
+                        .iter()
+                        .position(|v| v == &profile.scrollback_history);
 
                     let expanded_section = widget::settings::section()
                         .add(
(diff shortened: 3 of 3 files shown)


allen-munsch:
@smessmer looks like maybe the PR template is required, can you add it? the checklist and validate or affirm?

https://github.com/pop-os/cosmic-term/blob/master/.github/PULL_REQUEST_TEMPLATE.md

jklgrasso (approved):
Needs a review from @pop-os/ux team

jackpot51 (changes requested):
This PR does not even address #265, because it does not allow for setting the scrollback higher than 100,000 lines. Also, I would prefer it to be written without LLM use.

smessmer:
100,000 is a limit imposed by the alacritty library. There have been requests for alacritty to open that up to larger values previously, but the alacritty team rejected them with the argument that that could reduce performance and that performance is alacritty's top goal. So unless we can convince alacritty to open up that limit or unless we want to fork alacritty or switch to a different library, 100,000 is all we can do here.

I would much prefer an "unlimited" option myself and I would definitely set it to unlimited on my devices (even if it degrades performance by a bit), but given those constraints, 100,000 is at least better than the current default setting of cosmic-term.

About LLM use, I've written this PR with LLM help but have thoroughly read, tested and verified myself.

maria-komarova:
Thank you for taking the time to research the question and for thinking about possible improvements. We greatly appreciate it. 
After considering it the feature is not something we'd like to include in our Terminal settings. The impact for most users would be quite limited while adding unnecessary complexity to settings that we try to avoid when possible. This is part of COSMIC philosophy and we try to be mindful about what is included to keep things focused and beneficial for more people. That said if we see broader demand for this feature we can revisit it in the future.

jackpot51:
To follow up on what @maria-komarova said, I think we are okay to hard-code the maximum scrollback alacritty_terminal allows, 100,000 lines, until it can be made larger.

allen-munsch:
:face_with_diagonal_mouth:

i need to be able to configure the terminal lol

what if i don't want 100,000 lines of scrollback cause i'm running on a potato?

not being able to configure the terminal is a dealbreaker

why not have cosmic-term accept alacritty configs and pass them through so we can avoid this kind of back-and-forth?

if the goal is to simplify or mimic Apple's UI approach, that's your call: but configuring through the environment should be supported.

i tried several of these locations and it didn't work, not great UX or DX.

which part of the COSMIC philosophy does this fall under?

if you're going to force me into a new terminal on upgrade, please be more graceful with the constructive feedback.

https://alacritty.org/config-alacritty.html#location

```
Alacritty doesn't create the config file for you, but it looks for one in the following locations on UNIX systems:

    $XDG_CONFIG_HOME/alacritty/alacritty.toml
    $XDG_CONFIG_HOME/alacritty.toml
    $HOME/.config/alacritty/alacritty.toml
    $HOME/.alacritty.toml
    /etc/alacritty/alacritty.toml
```


Unless I missed something here, lemme knjow

jackpot51:
The part of alacritty we are using is `alacritty_terminal`, see https://docs.rs/alacritty_terminal for details. It does not include any of the alacritty GUI or config handling, that is all specific to cosmic-term and unrelated to alacritty. We can't just "accept" alacritty configs and "pass them through", as those parts of alacritty are not used in cosmic-term.

The concerns about performance would have to be backed up with evidence. Are you saying there is a system we'd expect COSMIC to run smoothly on with the current 10,000 lines scrollback but not with 100,000 lines scrollback? Do you have any benchmarking to demonstrate this?

allen-munsch:
okay sorry for misunderstanding, my bad

would you help me understand

https://alacritty.org/#About

> Alacritty is a modern terminal emulator that comes with sensible defaults, but allows for extensive configuration.


And then i see this:

```
09:36:37 jm@pop-os ~ → cat ~/.config/cosmic/com.system76.CosmicTerm/v1/profiles
{
    0: (
        name: "New profile",
        command: "",
        syntax_theme_dark: "COSMIC Dark",
        syntax_theme_light: "COSMIC Light",
        tab_title: "",
        working_directory: "",
        drain_on_exit: false,
// too complex?        scrollback_history: 100,
    ),
    1: (
        name: "New profile scroolbak",
        command: "",
        syntax_theme_dark: "COSMIC Dark",
        syntax_theme_light: "COSMIC Light",
        tab_title: "",
        working_directory: "",
        drain_on_exit: false,
// too complex?        scrollback_history: 100_000,
    ),
}
```

i dunno, just pointing it out, but maybe fork `alacritty_terminal` and call it `cosmic_emulator` since the philosophies appear to be irreconcilable?

jacobgkau:
Not wanting to implement a single feature in the COSMIC Term frontend that's implemented in the Alacritty frontend does not mean we need to fork the backend or that the "philosophies are irreconcilable." Alacritty provides the backend separately so that it can be used in other projects that are different from their own frontend (that's the entire reason the backend would be separated out).

Can you please answer Jeremy's question about what specific scenario 10,000 lines of scrollback would provide acceptable performance but 100,000 lines would not?

allen-munsch:
@jackpot51 @jacobgkau 

you ever go buy a bag of oranges? get home, and then find out its actually apples?

It's like buying a car that says "Ferrari engine inside" and finding out the steering wheel doesn't adjust and the seats are welded in place. Technically true about the engine, but the experience is totally different from what "Ferrari" implies.

don't advertise it as using alacritty in the README or even close to alacitty and then force your users into cosmic-term and give them the run around lol

just my 2 cents :shrug: 

alacritty's philosophy is "CUSTOMIZATION / CONFIGURATION"

COSMIC is not about configuration, that's cool, whatever, you do u

performance is kinda a strawman right? 

but 535MB RAM growth to 100K as a default seems kinda, i dunno a choice i guess

- https://github.com/alacritty/alacritty/issues/1236

an hypothetical example of maybe users would like to configure the terminal cause they have some reason XYZ?

so why are y'all coming at me about this?

its just the principle, should users be able to easily configure scroll back and all the other crap?

you're all clearly saying no, but like it was the first thing i looked at when i upgraded this junker that's duct taped together

its not some fluffy foo foo stuff if you have users asking for the number twiddler, right?

do you know what i mean?

---

@smessmer sorry didn't mean to hijack this PR

jacobgkau:
@allen-munsch The README simply states "built using [alacritty_terminal](https://docs.rs/alacritty_terminal)" because it is. Not implementing a particular setting in our frontend doesn't negate the fact that the backend crate is used in the code. You're choosing to read that technical statement in a source code repository as a stronger "advertisement" than it actually is.

You are perfectly welcome to use upstream Alacritty if you prefer the GUI and customizability of upstream Alacritty. You can change the default terminal app for COSMIC in Settings -> Applications -> Default Applications. As such, we are not "forcing our users into cosmic-term"-- it's just the preinstalled default.

Thank you for linking to the memory usage issue for some background about your concern. I'm assuming you have good intentions in advocating for this feature per our [code of conduct](https://github.com/pop-os/code-of-conduct); I'd ask that you return the courtesy and understand that we weren't "coming at you" or invoking strawmans by asking for data to back up what you're requesting.

allen-munsch:
yep overall good intentions

just was a frustrating experience is all

yah i've shifted back to gnome terminal, but hope to check in on cosmic-term from time to time

i like it, and appreciate the interaction, keep up the good work

gene1wood:
> To follow up on what @maria-komarova said, I think we are okay to hard-code the maximum scrollback alacritty_terminal allows, 100,000 lines, until it can be made larger.

@jackpot51 I've PRd that change in #836

gene1wood:
#836 is merged, released in [`epoch-1.0.15`](https://github.com/pop-os/cosmic-term/releases/tag/epoch-1.0.15) and available in the `cosmic-term` apt package version `1.0.15~1780432432~24.04~3832971`. So the new scrollback length is 100,000 though still not configurable.

PR #689, opened 2026-01-11
Setting for startup command

thorntonrose:
When the terminal starts up, it does not run the shell as a login shell (bash), which causes the shell to run .bashrc instead of .bash_profile. This PR adds a setting for a startup command. If the startup command is set, it will be run as the shell on startup. The format of the setting is "<command> <args>". Following the existing mechanism, the value of the setting is saved in ~/.config/cosmic/\<cosmic-term-dir>/v1/startup_command.

I saw in the code that there are undocumented flags for running a command (-e, --command, and --). Having to create a wrapper script and use it as a shim for cosmic-term is cumbersome, less user-friendly, and less visible than an app setting.

Note: Even though I know Rust, I don't use it regularly. So, I used GitHub Copilot to analyze and edit the code. I did review the changes to make sure they fit with existing patterns.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1535,7 +1535,7 @@ dependencies = [
 
 [[package]]
 name = "cosmic-term"
-version = "0.1.0"
+version = "0.2.1"
 dependencies = [
  "alacritty_terminal",
  "clap_lex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cosmic-term"
-version = "0.1.0"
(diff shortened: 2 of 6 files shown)


thorntonrose:
> You can already do this with profiles.

My understanding, from the little documentation I can find on profiles, it that the profile name has to be provided as command-line argument. Doing that requires that I create a wrapper script and use it as a shim, which is cumbersome and less visible than an app setting.

thorntonrose:
> You can already do this with profiles.

I see that now. Without docs, it's not clear how profiles work.

PR #690, opened 2026-01-11
Setting for xterm style middle-click paste

thorntonrose:
Many terminal applications (like PuTTY and Terminator) support right-click paste as an alternative to xterm style middle-click paste. This PR adds a setting for that, where on (default) = middle-click paste and off = right-click paste. Following the existing mechanism, the value of the setting is saved in ~/.config/cosmic/<cosmic-term-dir>/v1/xterm_middle_click.

Note: Even though I know Rust, I don't use it regularly. So, I used GitHub Copilot to analyze and edit the code. I did review the changes to make sure they fit with existing patterns.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1535,7 +1535,7 @@ dependencies = [
 
 [[package]]
 name = "cosmic-term"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


jackpot51:
I don't understand why this setting is necessary, you can already right click and select paste from the context menu.

PR #696, opened 2026-01-14
feat: add tab title editing feature

msenol:
## Summary
- Add tab title editing via context menu (right-click on tab)
- Add Ctrl+Shift+R keyboard shortcut for quick rename
- Add StringInterner for managed memory handling of tab titles
- Add `tab_rename_active` flag to properly disable terminals during rename
- Add i18n support for "rename-tab" in 40+ languages
- Add Escape key support to cancel rename dialog

## Implementation Details
- Tab titles are stored in `Terminal.tab_title_override`
- StringInterner manages `&'static str` allocations required by cosmic widget API
- The rename dialog appears as a centered overlay with semi-transparent backdrop
- Terminals are disabled during rename to prevent keyboard input leak

## Test plan
- [ ] Test tab rename via context menu
- [ ] Test tab rename via Ctrl+Shift+R keyboard shortcut
- [ ] Test rename in split panes (both horizontal and vertical)
- [ ] Test Escape key cancels rename
- [ ] Test empty title clears override
- [ ] Test keyboard input doesn't leak to terminal during rename

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,110 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Build & Development Commands
+
+```bash
+# Build (release)
+just build-release          # or: cargo build --release
+
+# Build (debug)
+just build-debug            # or: cargo build
+
+# Run with debug logging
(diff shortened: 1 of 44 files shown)


jackpot51 (changes requested):
Please only add new strings to the english translation file.

msenol:
I removed new strings from non-English translation files.

PR #698, opened 2026-01-16
feat: tab dnd

wash2:
This allows dragging to reorder tabs in the terminal window, but not moving them between windows. Does the drag threshold seem OK?


jackpot51 (dismissed):

wash2:
I think that this can be merged, but I guess it should be tested in case I missed any kind of issue like there was for cosmic-files or cosmic-edit.

garrettjwilke (dismissed):
drag and drop tabs works...threshold seems to be good, allowing the center of the tab as the "anchor point", but if you have 2 tabs open, and you begin to drag the tab to swap the positions, it feels like dragging one tab directly onto the other tab should just "swap them", but it requires you to move quite a distance in order to make them swap...this feels like it could just be my expectations...the tab placement indicator is visible and is easy to understand

testing with `cosmic-files` works great...tab drag works and dragging files between tabs also works

jackpot51 (dismissed):

atiq-cs:
Thank you @wash2

Issue #441, opened 2025-02-26
Support repositioning tabs via key bindings


jaypipes:
Would be great to "move" -- i.e. reposition -- tabs in cosmic-term using key combinations. For example, on gnome-terminal I use `Ctrl-Shift-PgUp` and `Ctrl-Shift-PgDn` to reposition tabs. Would be cool to have similar functionality in cosmic-term :)

PR #700, opened 2026-01-18
Add moving tabs left and right with key shortcuts

vacenty:
https://github.com/pop-os/cosmic-term/issues/441 This PR adds option to move tabs with left and right with Ctrl+Shift+Left/Right Arrow or by using buttons in view menu.


erik-balfe:
Reviewed against current `master` (cherry-pick conflicts in `src/key_bind.rs` and `src/main.rs` imports only; `menu.rs` + i18n merge cleanly).

**Why it does not merge as-is**
- `key_bind.rs` was refactored to delegate to `ShortcutsConfig` — new bindings belong in `shortcuts.rs`, not `key_bind.rs`.
- `Ctrl+Shift+ArrowLeft/Right` conflicts with existing split pane focus shortcuts on `master` (`PaneFocusLeft` / `PaneFocusRight`).
- `TabMoveLeft` / `TabMoveRight` also need wiring in `shortcuts.rs` (`KeyBindAction`, `action_label`, `shortcut_groups`, `fallback_shortcuts`) so they appear in View → Keyboard shortcuts and are user-remappable.

**What looks good**
- The `TabMoveLeft` / `TabMoveRight` handler logic using `tab_model.reorder()` is sound and matches the drag-and-drop path from #698.

**Suggestion for defaults**
Per #441, gnome-terminal-style `Ctrl+Shift+PageUp` / `Ctrl+Shift+PageDown` are free and avoid the pane-focus regression.

I'm preparing an updated PR rebased on `master` that incorporates this approach — happy to collaborate here or land separately. Thanks for getting this started!

vacenty:
I've rebased changes. I've changed default shortcuts to the ones suggested by @erik-balfe as they seem more standard than the ones that I'm used to. In case adding moving tabs left and right to menu feels like too much I'm open to removing it.

PR #721, opened 2026-02-05
Configurable keybinds

jackpot51:
Continuation of #674 for UX improvements
image(1)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2450,7 +2450,7 @@ impl Application for App {
                         self.shortcuts_config
                             .custom
                             .0
-                            .insert(binding, shortcuts::KeyBindAction::Unbind);
+                            .insert(binding, shortcuts::KeyBindAction::Disable);
                     }
                     shortcuts::BindingSource::Custom => {
                         self.shortcuts_config.custom.0.remove(&binding);
--- a/src/shortcuts.rs
+++ b/src/shortcuts.rs
@@ -56,7 +56,7 @@ impl Binding {
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub enum KeyBindAction {
-    Unbind,
+    Disable,
     ClearScrollback,
     Copy,
     CopyOrSigint,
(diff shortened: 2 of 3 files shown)


maria-komarova (dismissed):

src/shortcuts.rs
```diff
@@ -0,0 +1,557 @@
...
+            }
+        }
+
+        Some(KeyBind { modifiers, key })
+    }
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub enum KeyBindAction {
+    Unbind,
```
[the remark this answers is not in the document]
jackpot51: Done

src/main.rs
```diff
@@ -2149,6 +2429,59 @@ impl Application for App {
                 self.pane_model.set_focus(pane);
                 return self.update_focus();
             }
+            Message::ShortcutCaptureCancel => {
+                self.shortcut_capture = None;
+            }
+            Message::ShortcutCaptureStart(action) => {
+                self.shortcut_capture = Some(action);
```
scottwis: CONSIDER: Make this work similarly to cosmic-settings.

Cosmic settings will do this:

```
 iced_winit::platform_specific::commands::keyboard_shortcuts_inhibit::inhibit_shortcuts(true)
```

Before starting the capture.

See:

https://github.com/pop-os/cosmic-settings/blob/55b502dff39bf4be056c23888aaad185f9227eb2/cosmic-settings/src/pages/input/keyboard/shortcuts/common.rs#L442

This allows it to capture a key binding that is currently in use by cosmic-settings.

Right now if the user hits a key binding that is used by cosmic-comp, the key binding capture "just doesn't work". cosmic comp will consume the key and act on it. I think the current behavior here is reasonable for a first iteration, but this could be confusing for some users. The thing they are trying to do just "won't work" and they may not realize they have to go into another app, modify the settings, then come back to the terminal app.

It might be better to capture the keybinding and identify that it conflicts with a global key binding. Then a dialog could be displayed letting the user know there is a conflict with a global keybinding and that they have to use comsic-settings to disable it first. 

Though I don't think this needs to block the launch of the feature, and can probably be done as a follow up.  (enhancing it later shouldn't break backwards compatibility).  But you might want to consider this for a future enhancement.

jackpot51: It is not intended for a regular app to override window manager shortcuts

scottwis: I also don't think you need to take action on this PR.

leviport (approved):
This is looking great. I noticed the same [scroll jumping bug](https://github.com/pop-os/cosmic-term/pull/674#pullrequestreview-3752917136) as before, but I think we can just open a bug report for that one and fix it later. It's only a slight annoyance.

https://github.com/user-attachments/assets/5e18f7f0-c44b-421f-978e-67f700b55c38

PR #733, opened 2026-02-15
Add "Copy link" to context menu

chrisglass:
- [ x ] I have disclosed use of any AI generated code in my commit messages.
- [ x ] I understand these changes in full and will be able to respond to review comments.
- [ x ] My change is accurately described in the commit message.
- [ x ] My contribution is tested and working as described.
- [ X ] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

This branch adds a simple "Copy link" context menu, similar to the "open link" context menu.

The reason for this addition is that in my own workflow I often find myself copying links from my terminal (for instance, when discussing with people on IRC or such), and I find it inconvenient to "open link" then copy the now open browser's tab URL to copy it back to my notes.

I have considered the following implementations as well, which I am happy to further pursue if desired:

- Make the "Copy" action copy the underlying link if invoked when the user right-clicks a link. I opted against this because it makes things less discoverable, and would make the implementation of "copy" a bit more messy since it now sometimes needs to cares about the context menu's position (as far as I understand).
- Make the "launch URL by menu" action also copy the url in the clipboard. This is again quite "magic" and unexpected in my opinion.

Furthermore, I have considered to refactored the (almost identical) code between LaunchUulByMenu and the newly introduced CopyUrlByMenu Message handling, but decided against as it made the code (in my opinion) much less readable. Since it didn't hit the "rule of three" I opted to leave the duplication as-is.

(none of this code is generated by AI. It's mostly simple copying of existing behavior...)

EDIT: I took the liberty to add the French translation entry as well since I am a native speaker (and it's a trivial entry).

### DIFF
--- a/i18n/en/cosmic_term.ftl
+++ b/i18n/en/cosmic_term.ftl
@@ -119,6 +119,7 @@ clear-scrollback = Clear scrollback
 
 ## Open
 open-link = Open Link
+copy-link = Copy Link
 
 ## View
 view = View
--- a/i18n/fr/cosmic_term.ftl
+++ b/i18n/fr/cosmic_term.ftl
@@ -117,6 +117,7 @@ add-password = Ajouter un mot de passe
 password-input = Mot de passe
 password-input-description = Description
 open-link = Ouvrir le lien
+copy-link = Copier le lien
 add-another-keybinding = Ajouter un autre raccourci clavier
 cancel = Annuler
 close-window = Fermer la fenêtre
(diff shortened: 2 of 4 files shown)


leviport (approved):

leviport:
I requested a UX review since this includes an interface change (although it's a very minor one). Here is a screenshot of the new menu item:

Screenshot_2026-02-16_15-56-13

PR #742, opened 2026-02-20
feat: Add new keyboard-bindable action: clear and reset.

xward:
Hi !

Added QOL feature I miss from previous terminal in pop-os, a simple reset on terminal to flush everything. It looks like some other miss it too ! So here is my contribution.


jackpot51 (changes requested):
The PR template is mandatory and must be filled in. It is not correct to inject `reset` into the input.

PR #751, opened 2026-02-28
Fix "Copy link" lingering highlight bug

chrisglass:
Previously introduced "Copy Link" seems to have been missed when adding some changes  to "Open Link" which results in a lingering "underline" when copying links.

As a small drive by optimization I changes the match statement to match on both cases, since the treatment is the same. I could  have done this in the previous branch... but it escaped me at the time.

For some reason the file was not properly formatted according to `rust-fmt`, so some unrelated formatting changes snuck in. I can remove the auto format changes should they be on-purpose departures from `rust-fmt` standard, just let me know in the review.

This branch also removes an unnecessary import.

- [x ] I have disclosed use of any AI generated code in my commit messages.
- [ x] I understand these changes in full and will be able to respond to review comments.
- [ x] My change is accurately described in the commit message.
- [ x] My contribution is tested and working as described.
- [ x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,6 @@
 // Copyright 2023 System76 <info@system76.com>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use alacritty_terminal::tty::Options;
 use alacritty_terminal::{event::Event as TermEvent, term, term::color::Colors as TermColors, tty};
 use cosmic::iced::clipboard::dnd::DndAction;
 use cosmic::iced_core::keyboard::key::Named;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -1541,11 +1543,14 @@ impl App {
...
-                                    }),
-                                    drain_on_exit: startup_options.drain_on_exit || profile.drain_on_exit,
+                                    working_directory: startup_options.working_directory.or_else(
+                                        || {
+                                            (!profile.working_directory.is_empty())
+                                                .then(|| profile.working_directory.clone().into())
+                                        },
+                                    ),
+                                    drain_on_exit: startup_options.drain_on_exit
+                                        || profile.drain_on_exit,
```
chrisglass: This part is pure automatic formatting changes, created by rust-fmt

src/main.rs
```diff
@@ -1,7 +1,6 @@
 // Copyright 2023 System76 <info@system76.com>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use alacritty_terminal::tty::Options;
```
chrisglass: This was unused.

PR #757, opened 2026-03-09
Epoch 1.0.9 version update

jackpot51:
Generated by cosmic-epoch scripts/version-update.sh


hojjatabdollahi:
https://github.com/pop-os/cosmic-term/pull/776 should be merged before this.

jacobgkau (dismissed):
Dismissed this review (it seems the last commit just now finished building); it looks like this fixes the dGPU bug from master, which is good.

jacobgkau (approved):
Starting a custom profile, launching with a profile as default, all Edit functions, and all View functions (including all Settings) are working as expected.

PR #758, opened 2026-03-09
feat: enable ctrl+enter, shift+enter CSI u-escaping

bgub:

### DIFF
--- a/src/terminal_box.rs
+++ b/src/terminal_box.rs
@@ -952,6 +952,9 @@ where
                     Named::F10 => csi("21", "~", mod_no),
                     Named::F11 => csi("23", "~", mod_no),
                     Named::F12 => csi("24", "~", mod_no),
+                    Named::Enter if modifiers.shift() || modifiers.control() => {
+                        csi("13", "u", mod_no)
+                    }
                     _ => None,
                 };
                 if let Some(escape_code) = escape_code {


src/terminal_box.rs
```diff
@@ -952,6 +952,9 @@ where
                     Named::F10 => csi("21", "~", mod_no),
                     Named::F11 => csi("23", "~", mod_no),
                     Named::F12 => csi("24", "~", mod_no),
+                    Named::Enter if modifiers.shift() || modifiers.control() => {
+                        csi("13", "u", mod_no)
```
jackpot51: I'm curious if you have a source for this CSI? Both gnome-terminal and xterm output the same newline character on enter, ctrl+enter, and shift+enter.

bgub: @jackpot51 I'm using this as my main source: https://sw.kovidgoyal.net/kitty/keyboard-protocol/

I don't think either xterm or gnome-terminal implement the Kitty protocol (though the latter has an open PR).

bgub: @jackpot51 hmm, that's true. So the terminal would need to listen to that escape code and toggle some internal state to store whether it's in CSI U-mode?

I wonder how many apps rely on ctrl+enter or shift+enter working just like enter. I wonder if it could be fine just to emit the Kitty codes regardless of whether the app has explicitly opted in.

jackpot51: The kitty codes are supposed to be sent only when an app requests it, if they are sent otherwise the app may not be able to handle the input. Also. it looks like most keycodes are meant to be normalized after an app requests, not just ctrl+enter or shift+enter

PR #779, opened 2026-04-07
Fix binding to the pageup/pagedown keys

### LINKED ISSUE #727: Custom Shortcut with PageUp or PageDown not working
Hi everyone. I just update cosmic-term, using apt, and realize that now there is the option of "Keyboard shortcuts". 
It is great, but I realize a bug. I was trying to configure a shortcut for Next Tab and previous Tab, using PAGEDOWN and PAGEUP, but it is not working. 
When I successfully add "Ctrl + PAGEDOWN" to Next Tab, it doesn't work. I tried using "Ctrl + *" and it works correctly, so I think the problem is with PAGEDOWN and PAGEUP keys? Can you please check that? 
Btw whe I press Ctrl + PageUp or Ctrl + PageDown, it only prints "5~" in the terminal.
If you need more info I'm happy to help.

JellyWX:
Closes https://github.com/pop-os/cosmic-term/issues/727

Adds mapping from PageUp/PageDown keys in `key_to_string`. This allows users to bind to e.g ctrl+pgup/pgdown for switching between tabs.

Currently, hitting control+pgup/pgdown outputs `5~` in the terminal.

This is more of a short-term fix; it seems more sensible to update the type on the binding to be `Key` rather than `String` in the future, although I didn't look deeply into the implications of doing that. It seems that at a minimum, `SmolStr` would need to be ser/de which seems potentially problematic. Another idea might be to use a derive macro on `Named` to generate the mapping from string to variant (my Rust is fairly rusty at the moment so apologies if neither of those are operable).

## Testing

* Add a bind to Ctrl+PageUp/PageDown
* Validate that these keybinds now work for changing tabs
* Output some rubbish to the terminal (e.g `perl -e 'foreach (('a') x 100) { print rand() . "\n"; }'`) and validate that shift+pageup/pagedown still work for scrolling the terminal

## Checkboxes

- [ x ] I have disclosed use of any AI generated code in my commit messages.
- [ x ] I understand these changes in full and will be able to respond to review comments.
- [ x ] My change is accurately described in the commit message.
- [ x ] My contribution is tested and working as described.
- [ x ] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.


phlax:
fixed it for me!

jackpot51 (approved):

leviport (approved):

JellyWX:
Thank you!!

gene1wood:
This was released in [`1.0.14`](https://github.com/pop-os/cosmic-term/releases/tag/epoch-1.0.14).

PR #782, opened 2026-04-10
Fix clippy warnings

chrisglass:
This branches fixes all clippy warnings for this project. It is a lot of diff size but introduces no functional changes at all.
It is best reviewed per-commit (they all pass independently).

- A lot of mechanical changes as suggested by --fix (easy)
- The Message enum was very large because one of its variant has a very large type (DialogMessage is... 900+ bytes!). Passing it as a Box<> was an easy win (it is now only 200 bytes on the stack).
- One match statement was only ever matching a single case, so I added a comment and made it an `if` instead (saving a warning)
- A couple of functions trigger the "too many arguments" clippy lint, but that is a style preference in this case, so since the functions made sense to me I decided to ignore the warning for those particular cases.
- In order to hopefully make this excercise one-off, I added a check to the github actions to ensure clippy is run at least once before every merge.

This branch contains no LLM code, but the first commit results in the mechanical application of the command: `cargo clippy --fix --bin "cosmic-term" -p cosmic-term` . It is deterministic.

Edit: by boxing Config inside Message I can get the size of Message down to 64 bytes. I'm not sure it will have a noticeable impact on performance but ... it should be a single CPU cache line now. It will definitely save a lot of stack memory :)

### DIFF
--- a/.github/workflows/rust-checks.yml
+++ b/.github/workflows/rust-checks.yml
@@ -32,6 +32,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y pkg-config libxkbcommon-dev libfontconfig1-dev libfreetype6-dev libglvnd-dev libinput-dev libvulkan-dev libwayland-dev libx11-dev libxcursor-dev libxi-dev libxrandr-dev libasound2-dev libdbus-1-dev
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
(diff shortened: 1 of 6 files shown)


jacobgkau (dismissed):
All features & settings are still working as expected.

Not sure it was necessary to rename the existing rustfmt job, but it doesn't change the check itself. Engineering can weigh in on the new clippy and test jobs.

chrisglass:
> Not sure it was necessary to rename the existing rustfmt job

I reverted this, good catch! I initially tried to make them less tied to the tools since one step was a combined "clippy/tests" step but then changed my mind, and forgot the Rustfmt step :)

jacobgkau (approved):
Only change since my last approval was the CI job name.

jackpot51 (approved):

PR #797, opened 2026-04-30
feat(cli): add --window-size flag for initial window dimensions

HermeticOrmus:
## Why

`cosmic-term` has no equivalent of `xterm -geometry` or `alacritty --dimensions` — users triggering terminals from keyboard shortcuts or scripts can't pre-size the window, and UI resizes don't persist across launches. This is the ask in #771.

## What changed

`src/main.rs` gains a `--window-size <WxH>` flag (e.g. `--window-size 1280x720`), parsed from `clap_lex` alongside the existing `--working-directory` and `--no-daemon` handling.

When set, the value threads through libcosmic's `Settings::size(iced::Size::new(width, height))` before `cosmic::app::run`, so it takes effect on first paint. When unset, behavior is unchanged (libcosmic default of 1024x768).

Both lower- and uppercase separator (`x` / `X`) accepted. Floating-point dimensions accepted. Invalid input (missing separator, non-numeric, zero, negative) exits with a one-line error.

## How to test

```
cargo build --bin cosmic-term
./target/debug/cosmic-term --window-size 1280x720    # opens at the requested size
./target/debug/cosmic-term --window-size foo         # → "Invalid argument for --window-size: expected WxH, got \"foo\"", exit 1
./target/debug/cosmic-term --window-size             # → "Missing argument for --window-size", exit 1
./target/debug/cosmic-term --help                    # documents the new flag
cargo test --bin cosmic-term tests::                 # 6 new unit tests for parse_window_size
```

## Notes

- Naming follows the issue ("window-size") and the existing flag style. `--geometry` was considered but `xterm`'s geometry syntax (`WxH+X+Y`) implies position support, which would need a separate libcosmic API; staying with `--window-size` keeps the scope honest.
- Logical pixels (matching libcosmic's `iced::Size`), so the flag composes correctly with `COSMIC_SCALE`.
- 6 unit tests cover the parser: lower/uppercase separator, floats, missing separator, non-numeric, zero/negative.

Closes #771.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -98,6 +98,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let mut shell_args = Vec::new();
     let mut daemonize = true;
     let mut working_directory = None;
+    let mut window_size = None;
     // Parse the arguments using clap_lex
     while let Some(arg) = raw_args.next_os(&mut cursor) {
         match arg.to_str() {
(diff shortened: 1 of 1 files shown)


jackpot51:
The PR template is mandatory

PR #800, opened 2026-05-01
feat(font): add disable_ligatures toggle in advanced font settings

### LINKED ISSUE #735: Support toggling (in particular, disabling) ligatures.
I've been using `cosmic-term` recently and immediately noticed that it renders ligatures. I'd really prefer to disable them. However, I can't seem to find a way to do this. IMO, this should be possible from the GUI. Can ligature rendering be configured? If so, how? If not, what do folks think about providing a toggle for this in the `View > Settings > Font > Advanced font settings` UI or someplace similar?

Thanks!

HermeticOrmus:
Closes #735.

Adds `Config.disable_ligatures` (default: `false`). When enabled, disables the OpenType `liga`, `clig`, and `calt` features on the terminal's default font attrs.

The toggle is exposed in **View > Settings > Font > Advanced font settings**, alongside the existing `Make bold text brighter` toggle, matching the location requested in the issue.

Changes propagate live to existing terminals through `Terminal::set_config`, so toggling does not require restarting tabs.

## Files

- `i18n/en/cosmic_term.ftl` — `disable-ligatures = Disable ligatures`
- `src/config.rs` — `disable_ligatures: bool` field, `#[serde(default)]`
- `src/main.rs` — `Message::DisableLigatures`, settings row, update handler
- `src/terminal.rs` — `no_ligature_features()` helper, applied in `new` and on live config change

## Verification

- `cargo check` passes
- Default behaviour unchanged when flag is `false`
- Toggle applies live without tab restart

### DIFF
--- a/i18n/en/cosmic_term.ftl
+++ b/i18n/en/cosmic_term.ftl
@@ -54,6 +54,7 @@ default-font-weight = Normal font weight
 default-dim-font-weight = Dim font weight
 default-bold-font-weight = Bold font weight
 use-bright-bold = Make bold text brighter
+disable-ligatures = Disable ligatures
 
 ### Splits
 splits = Splits
--- a/src/config.rs
+++ b/src/config.rs
@@ -238,6 +238,8 @@ pub struct Config {
     pub default_profile: Option<ProfileId>,
     #[serde(default)]
     pub shortcuts_custom: Shortcuts,
+    #[serde(default)]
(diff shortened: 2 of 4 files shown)


jackpot51:
The PR template is mandatory

vertigo-z:
@HermeticOrmus I want this feature. Would you be able to resubmit the PR with the proper template? Otherwise I will open one myself and attribute you.

PR #806, opened 2026-05-08
bound the event channel to prevent unbounded RSS growth

### LINKED ISSUE #805: Memory leak: cosmic-term consumes 254GB RSS in long sessions, killed by OOM
During long terminal sessions I noticed that my machine would become 
completely unresponsive. Checking the logs after the fact, I found that 
cosmic-term had been killed by the OOM killer after consuming nearly all 
available RAM (254GB on a 256GB system). I traced this back to the 
unbounded handling of terminal events in EventProxy.

## Environment

- Pop!_OS 24.04
- cosmic-term 1.0.11
- Kernel 6.18.7-76061807-generic
- 256GB RAM 

# Logs: 

```
May 06 20:53:31 LocalMachine kernel: [   4901]  1000  4901 69641705 63735488 63734710      778         0 554528768  5237984             0 cosmic-term

May 06 20:53:31 LocalMachine kernel: oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/session-3.scope,task=cosmic-term,pid=4901,uid=1000

May 06 20:53:31 LocalMachine kernel: Out of memory: Killed process 4901 (cosmic-term) total-vm:278566820kB, anon-rss:254938840kB, file-rss:3112kB, shmem-rss:0kB, UID:1000 pgtables:541532kB oom_score_adj:0

May 06 20:53:31 LocalMachine systemd[1]: session-3.scope: A process of this unit has been killed by the OOM killer.

May 06 20:53:34 LocalMachine kernel: oom_reaper: reaped process 4901 (cosmic-term), now anon-rss:0kB, file-rss:404kB, shmem-rss:0kB
```

anon-rss 254.9GB (resident heap), total-vm 278.5GB

The anon-rss figure shows the memory was in private heap allocations 
within cosmic-term itself, consistent with an unbounded queue 
accumulating events over time.

marshallhumble:
- [ X] I have disclosed use of any AI generated code in my commit messages.
- [X ] I understand these changes in full and will be able to respond to review comments.
- [ X ] My change is accurately described in the commit message.
- [ X] My contribution is tested and working as described.
- [ X] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

Fixes #805 

This changes the  mpsc::unbounded_channel() to a bound one with a cap of 1024 messages. I used 1024 here since that should be at max 100MB. I have built and tested this locally and the terminal is "locked up" but the machine is not. This to me seem to fix the issue, previously the terminal consumed all RAM on the system. 

AI Usage: Claude Code was used in the initial troubleshooting of the system problem and for code review. 

Test: 

run ```yes | head -c 1G``` in terminal to create messages. This only causes cosmic-term to become unresponsive and not the system, as seen previously. 

Btop output with local changes applied to cosmic-term: 

image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -438,7 +438,7 @@ pub enum Message {
     TabNext,
     TabPrev,
     TermEvent(pane_grid::Pane, segmented_button::Entity, TermEvent),
-    TermEventTx(mpsc::UnboundedSender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>),
+    TermEventTx(mpsc::Sender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>),
     ToggleFullscreen,
     ToggleContextPage(ContextPage),
     UpdateDefaultProfile((bool, ProfileId)),
@@ -503,7 +503,7 @@ pub struct App {
     find_search_id: widget::Id,
     find_search_value: String,
     term_event_tx_opt:
-        Option<mpsc::UnboundedSender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>>,
+        Option<mpsc::Sender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>>,
     startup_options: Option<tty::Options>,
     term_config: term::Config,
     color_scheme_errors: Vec<String>,
@@ -3623,7 +3623,7 @@ impl Application for App {
                 stream::channel(
                     100,
                     |mut output: iced::futures::channel::mpsc::Sender<Message>| async move {
-                        let (event_tx, mut event_rx) = mpsc::unbounded_channel();
+                        let (event_tx, mut event_rx) = mpsc::channel(1024);
                         output.send(Message::TermEventTx(event_tx)).await.unwrap();
 
                         while let Some((pane, entity, event)) = event_rx.recv().await {
--- a/src/main.rs
+++ b/src/main.rs
@@ -502,8 +502,7 @@ pub struct App {
     find: bool,
     find_search_id: widget::Id,
     find_search_value: String,
-    term_event_tx_opt:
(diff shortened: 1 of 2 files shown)


chrisglass:
I have tried this branch and can confirm it does indeed solve the problem. 
Thanks!

One small request: could you please add comments next to the changes (Sender vs. UnboundSender, channel creation) explaining why those are done? For example, a one-line comment explaining that 1024 messages is about 100Mb and prevents the OS from locking because of OOM would be nice (avoids future travelers from asking themselves why that specific number).

wash2 (dismissed):

marshallhumble:
Added comments per @chrisglass 's suggestion ran clippy/cargo fmt no code changes

wash2 (approved):

jacobgkau (approved):
I'm seeing COSMIC Term's scrollback limited to just over 10,000 lines (10,028 on a physical machine, 10,018 in a VM) with or without this change. The example command `yes | head -c 1G` to replicate the issue this purports to fix did not work for me (it just completed successfully, without a noticeable bump in RAM usage, due to that scrollback limit), both on a physical machine and in a VM. (I tried disabling graphics acceleration in the VM, and saw the RAM usage get a little higher at first, but still didn't see it grow indefinitely.) From what I understand based on the code comments, this issue would occur if the terminal's not able to display output quickly enough to keep up with what's being output, so it would be specific to an environment where that would occur.

If engineering agrees with the change, I have no reason to oppose it from the QA side, since I'm not seeing any difference in behavior with it. Regression testing of all other functionality (profiles, settings, keyboard shortcuts, passwords, etc) looks fine.

Issue #781, opened 2026-04-10
Mouse wheel scroll events heavily attenuated — requires fast scrolling to register


dkleehammer:
### Description

Mouse wheel scroll events are severely attenuated in cosmic-term. A single notch of the mouse wheel produces zero input — you have to spin the wheel very fast to get any scroll response at all. This affects any application that relies on terminal mouse/scroll escape sequences (tmux, less, vim, etc.).
Touchpad two-finger scrolling works perfectly in all cases.
The terminal's own scrollback buffer scrolls fine with the mouse wheel (the widget-level scroll handler). The bug is specifically in the generation of escape sequences sent to the PTY when the mouse wheel is used.

### Steps to Reproduce

1. Open cosmic-term
2. Run `cat -v`
3. Slowly scroll the mouse wheel — one notch at a time
4. Expected: Each notch should produce `^[[A` or `^[[B` (or mouse button 4/5 escape sequences if mouse reporting is enabled)
5. Actual: Nothing is output. No escape sequence is generated.
6. Now spin the mouse wheel very fast — eventually `^[[A` will appear, but only after aggressive scrolling.

### Comparison with touchpad

1. Run `cat -v`
2. Do a small two-finger scroll on the touchpad
3. Escape sequences are generated immediately and responsively

### Demonstrating the impact in tmux

1. Start tmux (even with an empty config file — `tmux -f /dev/null`)
2. Generate enough output to fill the scrollback (e.g., `seq 1000`)
3. Try scrolling back with the mouse wheel — requires very fast spinning to move at all
4. Try scrolling back with the touchpad — works perfectly

Mouse wheel scrolling works correctly in all other COSMIC applications (cosmic-files, cosmic-settings, etc.), so the compositor and libinput are delivering correct scroll events.

### Probable Cause
Super guessing here and so is Claude... the scroll event handling in `terminal_box.rs` likely converts pixel-based scroll deltas (from iced/Wayland) into discrete terminal escape sequences. On Wayland, mouse wheel events may arrive as`ScrollDelta::Pixels` rather than `ScrollDelta::Lines`. A single discrete wheel notch produces a small pixel delta (e.g., ~15px), which when divided by the cell height (~20px at font size 16) produces a fractional value less than 1.0 — rounding to zero and generating no event.

Touchpad scrolling works because it generates many continuous small events that accumulate, while a discrete mouse wheel notch is a single event that gets discarded.

After walking through a bunch of this with Claude, it thinks fix is likely one of:

[...]

Issue #385, opened 2024-12-09
[Bug] Can't scroll using the mouse in pager software (less and more)


nastynick64:
I'm using Cosmic Terminal on KDE Plasma 6.2.4 on Arch Linux

I've noticed this issue where whenever I try to read a man page I can't seem to scroll with the mouse, scrolling with up and down arrow still works tho. 

I've also tried piping `history` through `less`, and also `more`. It doesn't work there either. 

I remember on some older version this wasn't an issue, I don't remember which.

dkleehammer:
Having the same issue, I am able to scroll if I spin the mouse wheel very fast... but this is not a viable option.

PR #811, opened 2026-05-12
Fix fractional wheel scrolling in scrollback and alternate screen

jerezereh:
Addresses issues #781 (commit 30738dbe34f087535d670fe4e6d1d34edfa89817) and #385 (commit 04683470b572abdd581dc56a234abec5c9419551).

Wheel scrolling previously cast float inputs to ints, dropping all fractional scrolling that occurs on most modern, high-resolution scroll wheels. Commit 30738dbe34f087535d670fe4e6d1d34edfa89817 addresses this by keeping scroll delta as float, and saving remainders to state.

Additionally noticed while testing that scroll events were not received in ALT_SCREEN mode (`less`, `vi`, `man`, etc.). Fixes in commit 04683470b572abdd581dc56a234abec5c9419551 address this by allowing non-SGR mouse mode wheel events, sending mouse wheel or arrow key events depending on TermMode::APP_CURSOR, and saving the fractional mouse wheel events (as in the previous commit).

Tests have been added to both `terminal_box.rs` and `mouse_reporter.rs`.

---
---
Disclaimer: As I am new to Rust, AI agent Codex was used for reference and to review code. If any changes here don't look right, I would be happy to receive feedback or guidance.


pforhan:
Not sure if anyone has looked at this, but I wonder if they're hesitant about using floats here.  I'm no rust guy either, so not sure how impactful the change is...

jerezereh:
The data from the mouse wheel is actually already coming in as a float and is being cast to an int, which is part of the precision loss issue for the scrollback. Unless there's some larger concern about using floats in Rust I'm unaware of, that change should be fairly innocuous.

PR #818, opened 2026-05-18
Revert "bound the event channel to prevent unbounded RSS growth"

mmstick:
Reverts pop-os/cosmic-term#806

Alacritty-term's spawned event loop holds a lease to `Terminal::term`'s mutex until it is finished sending events back to the app. This causes a deadlock at [src/terminal.rs:760 in Terminal::update](https://github.com/pop-os/cosmic-term/blob/master/src/terminal.rs#L760) when the channel is full. For now, it would be best to revert until we refactor this. We may want to avoid sharing a mutex with the event loop.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -438,7 +438,7 @@ pub enum Message {
     TabNext,
     TabPrev,
     TermEvent(pane_grid::Pane, segmented_button::Entity, TermEvent),
-    TermEventTx(mpsc::Sender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>),
+    TermEventTx(mpsc::UnboundedSender<(pane_grid::Pane, segmented_button::Entity, TermEvent)>),
     ToggleFullscreen,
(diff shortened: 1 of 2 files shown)


jacobgkau:
The initial PR was clear about the deadlock when the channel's full:

> This changes the mpsc::unbounded_channel() to a bound one with a cap of 1024 messages. I used 1024 here since that should be at max 100MB. I have built and tested this locally and the terminal is "locked up" but the machine is not. This to me seem to fix the issue, previously the terminal consumed all RAM on the system.

I don't care one way or the other how this is handled, but wanted to point out that this got an engineering approval before. Have we had the deadlock actually happen for people (who weren't previously affected by the OOM issue) since the PR was merged?

mmstick:
Yes, https://chat.pop-os.org/pop-os/pl/64gbyi9y4t8ixqskjc9mee56ja

jacobgkau (approved):

jackpot51 (approved):

Issue #813, opened 2026-05-14
Feature request: Implement full OSC sequence support from alacritty_terminal


gblach:
## Description

  cosmic-term is built on `alacritty_terminal`, which already parses a defined set of OSC sequences. However, several of these are not wired through to cosmic-term's UI and renderer layers, meaning they are silently ignored at runtime.

  ## OSC sequences

  | OSC | Reset | Description | Test command | Reset command | Status |
  |-----|-------|-------------|--------------|---------------|--------|
  | `0` | - | Set window title and icon name | `printf '\033]0;test title\007'` | - | working |
  | `2` | - | Set window title | `printf '\033]2;test title\007'` | - | working |
  | `4` | `104` | Set/query palette color | `printf '\033]4;1;#00ff00\007'` | `printf '\033]104;1\007'` | **broken** |
  | `8` | - | Hyperlinks | `printf '\033]8;;https://example.com\033\\click here\033]8;;\033\\'` | - | **broken** |
  | `10` | `110` | Default foreground color | `printf '\033]10;#ff0000\007'` | `printf '\033]110;\007'` | **broken** |
  | `11` | `111` | Default background color | `printf '\033]11;#800080\007'` | `printf '\033]111;\007'` | **broken** |
  | `12` | `112` | Cursor color | `printf '\033]12;#ff0000\007'` | `printf '\033]112;\007'` | working |
  | `50` | - | Cursor shape | `printf '\033]50;CursorShape=1\007'` | - | working |
  | `52` | - | Clipboard write | `printf '\033]52;c;%s\007' $(printf 'hello' \| base64)` | - | working |

  ## Steps to reproduce (broken)

  ```sh
  # Palette color - no effect
  printf '\033]4;1;#00ff00\007'
  printf '\033[31mthis should be green if OSC 4 worked\033[0m\n'

  # Hyperlinks - no effect
  printf '\033]8;;https://example.com\033\\click here\033]8;;\033\\'

  # Foreground color - no effect
  printf '\033]10;#cdd6f4\007'

  # Background color - no effect
  printf '\033]11;#800080\007'
  ```

  ## Expected behavior

  Each sequence produces the described effect immediately. Color changes persist until explicitly reset or the terminal closes. Query sequences (`?`) respond with the current value.

  ## Root cause

  `alacritty_terminal` handles VT parsing and stores the resulting state, but cosmic-term's rendering and window management layers need to:

[...]

PR #853, opened 2026-06-19
fix: report effective OSC 11 background

Xanewok:
---
This makes OSC 11 return the same background color that is being rendered, i.e. it accounts for the current `WINDOW_BG_COLOR` that is set by theme.

This is meant as a simple fix for now without trying to unify/refactor the rest of the code, so this duplicates the simple code path and copies the same comment for easier greppability; please let me know if you'd like to see the unification as part of this PR.

Part of #813 

For example, Vim uses it to detect the themes, and so here's an example rendering of Cargo.toml before (notice the bright pink, which shouldn't be a foreground color on a light theme):
before

and after:
after

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3066,7 +3066,7 @@ impl Application for App {
                             && let Some(terminal) = tab_model.data::<Mutex<Terminal>>(entity)
                         {
                             let terminal = terminal.lock().unwrap();
-                            let rgb = terminal.colors()[index].unwrap_or_default();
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jklgrasso:
This needs to be rebased. Once done, I will have CI rebuild. Assuming that works (it built locally, so it should be fine), I will approve

Xanewok:
Rebased :pray:

jklgrasso (approved):

PR #864, opened 2026-07-09
fix: only PaneClicked events change focus

RipleyTom:
This changes the focus logic so that clicking the title bar(to move the window around) or resize buttons doesn't automatically unfocus the current pane. All the pane focus logic goes through PaneClicked=>update_focus.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2582,7 +2582,8 @@ impl Application for App {
             }
             Message::PaneClicked(pane) => {
                 self.pane_model.set_focus(pane);
-                return self.update_title(Some(pane));
+                let title_task = self.update_title(Some(pane));
+                return Task::batch([self.update_focus(), title_task]);
             }
             Message::PaneSplit(axis) => {
                 let result = self.pane_model.panes.split(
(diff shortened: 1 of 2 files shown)


git-f0x:
I seems `cosmic-edit` also has the text box getting unfocused in similar cases.

jackpot51 (approved):

RipleyTom:
I've made a more generic libcosmic fix that fixes all applications focus issues with libcosmic: https://github.com/pop-os/libcosmic/pull/1342 . If it's accepted I'll just close this.
The libcosmic fix is a bit better as this only affects the panes but not for example the search edit, which loses focus if you move the window.

RipleyTom:
Nvm about the generic libcosmic thing, it's scrapped.

LinuxGratos:
This is the fix we needed ! Thank you
It was bothering me that the terminal loses focus
Too bad it won't be in 1.3.0 :cry:

dania83:
I confirm. I have a problem of terminal hung when I use maximize button. With this commit the problem is resolved

jacobgkau (approved):
Fixes keyboard focus after clicking/dragging the titlebar. All menu options are still working as expected.

PR #881, opened 2026-07-22
feat: Dropdown terminal

rnbwdsh:
* Add --drop-down CLI flag, that adds new behavior and keeps all the old behavior if not used
* Behavior on first launch: create a surface, add a signal handler, render the terminal into that overlay surface
* Behavior on later launches: find the process that was launched with "--drop-down" and send a signal to it (currently via spawning a process) Handle USR1 signals for toggling dropdown
* Custom right click context menu in dropdown mode, as maximizing makes no sense
* Added setting: A secondary show-headers-dropdown that's only visible in drop down mode
* Added setting: The height of the dropdown in px, I'd like to do relative but to my understanding wayland doesn't trivially allow that
* Bug byepass in iced: When creating the overlay without any delay, the window is transparent but not "frosted"

This is hugely based on pop-os#673 and just rebased and refactored, as various menu things like the header and the context menu were broken for me, and the pid file thing was overly complicated and introduced too many dependencies.

I think this MR is relevant, because people are already vibe coding a non-working version of this: https://github.com/M0Rf30/cosmic-ext-quake-terminal

Here is a screenshot of how it looks for me:
Screenshot_2026-07-22_11-44-40

And here one without header + with settings open:
Screenshot_2026-07-22_12-11-27

This was partially created with the help of AI.

### DIFF
--- a/src/dropdown.rs
+++ b/src/dropdown.rs
@@ -1,0 +1,160 @@
+use cosmic::Application;
+use cosmic::iced::{
+    self, Subscription,
+    futures::SinkExt,
+    platform_specific::runtime::wayland::layer_surface::SctkLayerSurfaceSettings,
+    platform_specific::shell::commands::layer_surface::{
+        Anchor, KeyboardInteractivity, Layer, destroy_layer_surface, get_layer_surface,
+    },
+    stream, window,
+};
+use cosmic::style;
+use cosmic::widget;
+use std::any::TypeId;
+
+use crate::App;
+use crate::Message;
+
+/// Scan /proc for an existing cosmic-term process with --drop-down flag.
+pub fn find_existing_dropdown_pid() -> Option<u32> {
+    let current_pid = std::process::id();
+    let proc_dir = match std::fs::read_dir("/proc") {
+        Ok(d) => d,
+        Err(_) => return None,
+    };
+    for entry in proc_dir.flatten() {
+        if let Some(pid) = entry
+            .file_name()
+            .to_string_lossy()
+            .parse::<u32>()
+            .ok()
+            .filter(|&p| p != current_pid)
+            .and_then(|pid| {
+                std::fs::read(entry.path().join("cmdline"))
+                    .ok()
+                    .filter(|cmdline| {
+                        let args: Vec<&[u8]> = cmdline.split(|b| *b == 0).collect();
+                        args.iter().any(|a| {
+                            a.ends_with(b"cosmic-term")
+                                || a.ends_with(b"cosmic-term\n")
+                                || a == b"cosmic-term"
+                        }) && args.iter().any(|a| *a == b"--drop-down")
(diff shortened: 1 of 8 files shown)


src/dropdown.rs
```diff
@@ -0,0 +1,160 @@
...
+};
+use cosmic::style;
+use cosmic::widget;
+use std::any::TypeId;
+
+use crate::App;
+use crate::Message;
+
+/// Scan /proc for an existing cosmic-term process with --drop-down flag.
+pub fn find_existing_dropdown_pid() -> Option<u32> {
```
rnbwdsh, line 19: note: this could also be done with a PID file, but imo that's more complicated. it would look like this https://github.com/pop-os/cosmic-term/pull/673/changes#diff-d97a84a0cb3e75d36453a8924753487d4bed7458a74ad58da9189a5b7082e1ef

src/main.rs
```diff
@@ -159,6 +165,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    // Drop-down mode single-instance handling:
+    // scan /proc for an existing cosmic-term --drop-down process.
+    if drop_down {
+        // Look for a running cosmic-term --drop-down process.
+        if let Some(pid) = dropdown::find_existing_dropdown_pid() {
+            log::info!(
+                "Found existing drop-down instance (PID {}), sending toggle signal and exiting",
+                pid
+            );
+            let _ = std::process::Command::new("kill")
```
rnbwdsh, line 177: this could also be done by importing libc or the nix crate, but it would add an unsafe block. on the other hand, this variant requires the "kill" binary

src/main.rs
```diff
@@ -1899,12 +1964,23 @@ impl Application for App {
...
         app.set_curr_font_weights_and_stretches();
-        let command = Task::batch([app.update_config(), app.update_title(None)]);
 
-        (app, command)
+        let mut commands = vec![app.update_config(), app.update_title(None)];
+        // on initial drop-down mode launch, if we instantly toggle the drop-down, the transparency is not yet frosted
+        // it's fine on later launches and delaying it just 1ms fixes it too
+        if app.is_dropdown {
+            commands.push(cosmic::task::future(async {
+                tokio::time::sleep(std::time::Duration::from_millis(1)).await;
```
rnbwdsh, line 1985: if i don't do this, the window is transparent but not frosted. i think that's a bug in iced?

src/terminal_box.rs
```diff
@@ -543,7 +543,7 @@ where
                     fn fill<Renderer: renderer::Renderer>(
                         &mut self,
                         renderer: &mut Renderer,
-                        is_focused: bool,
+                        _is_focused: bool,
```
sriharshachilakapati, line 546: Why this change? I don't see this being used inside the function though..

Issue #878, opened 2026-07-20
Color schemes are broken with frosted glass


snaggen:
When frosted glass is used, the frosted glass global translucency is used and the terminal background looks the same as the header bar.  This is not consistent with other applications, like comsic-monitor where the information cards have a much darker (I use a dark theme) background with less translucency. For a terminal, the terminal box should be considered to be the information carrying part and should hence should have a higher contrast much like the information cards in cosmic-monitor. 
Also, the color schemes for the cosmic-terminal are defined with different background, some with high contrast, some with lower contrast. So, the different color schemes will behave quite different in regards to usability with translucency. Hence, the terminal needs to be able to specify it's translucency and not just rely on the default. 

I have also noticed, that the dark tint for non focused panes are gone, making it harder to see which pane that is active. I saw that this was removed since it didn't match the header bar, which was a false premise, since the terminal box should not have the same translucency as the header bar in the first place.

PR #889, opened 2026-07-27
Adjustable translucensy

snaggen:
I have, with the help of Claude Code, made a fix for #878 . Now the opacity configuration is now shown for frosted glass effect, but defaults to the opacity value for the Primary layer and not the Background layer which it was previously. I think Primary might be a bit low, but it is a lot better than using the same as the Background layer. I also restored the non-focus tint. 

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@
 
 use cosmic::{
     cosmic_config::{self, CosmicConfigEntry, cosmic_config_derive::CosmicConfigEntry},
+    cosmic_theme::BlurStrength,
     theme,
 };
 use cosmic_text::{Metrics, Stretch, Weight};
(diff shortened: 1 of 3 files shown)


snaggen:
Did some cleanup. And made frosted glass default opacity use BlurStrength::VeryLow by default since that seems to adjust after the users theme.

snaggen:
I wonder if it is better to add a separate slider and a blur_opacity setting, to be used only for frosted glass, where the slider have distinct steps matching the BlurStrenght enums. That way the blur opacity will always be relative the setting in cosmic-settings. Right now, it is relative to that until you change the value, then it gets fixed. 
What do you think @jackpot51 ?

snaggen:
I implemented a separate slider and config to keep the blur opacity relative the main blur settings.

Issue #363, opened 2024-10-24
Subpar font hinting on low-DPI displays


LunarLambda:
The default hinting used by cosmic-term/cosmic-text looks doesn't look as good as most other apps/terminals, even at larger font sizes (15pt/20px in my case). Edges are overly sharp and there's no subpixel rendering. It's particularly noticeable on `o`s, `e`s, as well as on brightly colored text, where stroke widths can end up looking slightly uneven.

Example: Top: `cosmic-term`, Bottom: `alacritty`
![image](https://github.com/user-attachments/assets/6b1fb979-b2a9-4f7f-9f3a-b9e3542e6f8e)

Zoomed in:
![image](https://github.com/user-attachments/assets/fb5f00a9-ad59-42a1-a72f-6f08262e3ba5)

`wezterm` with subpixel rendering (`freetype_render_target = 'HorizontalLcd'`) gets pixel-for-pixel identical looking results to `alacritty` (minus differences in colorschemes), as does `gnome-terminal`.

charliechensz:
Confirming this issue is still present in 1.5.0, with a concrete data point: **cosmic-term 1.5.0 changed its render stack** (1.3.0 → 1.5.0: from system fontconfig to an embedded one, using the skrifa/swash shaping stack), and font rendering quality visibly regressed — grainy/fuzzy glyphs at typical sizes, notably lacking subpixel antialiasing, while kitty on the same system (freetype + subpixel AA) renders cleanly.

Environment: Pop!_OS 24.04, cosmic-term 1.5.0, NVIDIA RTX 4060, typical desktop monitor (~96 DPI).

Observations:
- Low-DPI displays are where this hurts most — exactly the "overly sharp edges, no subpixel rendering" described in the screenshots here.
- The render-stack change in 1.5 did fix a separate issue (glyph drift in Neovim), so it is not a pure regression — but font appearance is a real downgrade.
- Workaround for now: I switched to kitty. But cosmic-term is the default terminal, so it would be great to see this fixed.

Suggestions:
1. Enable subpixel antialiasing (or at least better hinting) — critical on ≤96 DPI displays.
2. Check whether the embedded fontconfig config disables hinting/subpixel by default.
3. Expose an AA mode + hinting strength option in advanced font settings.

Happy to provide side-by-side screenshots (cosmic-term vs kitty) if helpful.

charliechensz:
[...]

PR #895, opened 2026-08-03
fix: display all monospace fonts from the system

hasanmohammad6618:


JackNWhite:
I tried this out. It's a good first step and did give me access to Terminus, which is a dealbreaker for me, but it didn't render correctly. The text was extremely fuzzy, even after playing with the different Terminus nerdfont sizes. Zed seems to have fixed this, and IIRC it's based on COSMIC Text, so I don't know whether the problem here is simply the enforcement of antialiasing, or whether it's the bizarre scaling rules which mean that fonts in COSMIC tend to be much, much smaller than in other environments.

Anyway, it's a good fix, but I wanted to give the feedback, because from the comment in the code you removed, there was clearly a reason it was there in the first place.

There is, presumably, a way around this though, because Zed does handle it.

hasanmohammad6618:
> I tried this out. It's a good first step and did give me access to Terminus, which is a dealbreaker for me, but it didn't render correctly. The text was extremely fuzzy, even after playing with the different Terminus nerdfont sizes. Zed seems to have fixed this, and IIRC it's based on COSMIC Text, so I don't know whether the problem here is simply the enforcement of antialiasing, or whether it's the bizarre scaling rules which mean that fonts in COSMIC tend to be much, much smaller than in other environments.
> 
> Anyway, it's a good fix, but I wanted to give the feedback, because from the comment in the code you removed, there was clearly a reason it was there in the first place.
> 
> There is, presumably, a way around this though, because Zed does handle it.

Maybe because of this -> https://github.com/pop-os/cosmic-term/issues/363

PR #915, opened 2026-09-08
feat: use libcosmic's context_menu widget for the terminal context menu

hojjatabdollahi:
Drops the custom implementation for wayland popup context menu, and uses `libcosmic`'s `context_menu` widget.

Depends on
- [x] https://github.com/pop-os/libcosmic/pull/1416

The second commit, removes `cosmic-files` as a dependency and instead uses `libcosmic`'s rfd feature. If that's not desired, then I can drop that commit and wait for  https://github.com/pop-os/cosmic-files/pull/2052.

____

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,7 +115,7 @@ checksum = "35f0f96ce78e38c3dc6d8948aa8163d06385be74000f3c7a95bf1eef35d3ea32"
 dependencies = [
  "cipher",
  "cpubits",
- "cpufeatures 0.3.1",
+ "cpufeatures",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


jklgrasso (commented):
https://github.com/pop-os/cosmic-term/pull/917#pullrequestreview-5213045616 :

> The "Show header" option works, but the toggle is not visible.
> 
> This PR:
> Image
> 
> Current release:
> Image

jklgrasso:
Also noticed the header context menu will take mouse clicks even when the cursor is not in the context menu. The context menu for the terminal itself does not do this

https://github.com/user-attachments/assets/d048eabd-ad0a-42f1-be24-433c53f1fbbe

Edit: This is not a regression in behavior

hojjatabdollahi:
The show header toggle was abnormal. No other context menu has a toggle, they all have a checkmark, and that's what libcosmic provides.

~But if it's needed, we can create a custom menu item to support elements in the menu.~ actually that would be pretty hard to handle arbitrary elements with their messages, so probably we can recreate the previous UI exactly with the same UX (clicking the toggler closes the menu instead of showing a toggle animation).

jklgrasso:
> The show header toggle was abnormal. No other context menu has a toggle, they all have a checkmark, and that's what libcosmic provides.

I agree that it was abnormal. Switching to a checkmark would probably be good. It would match similar UI and provide a more consistent UX.

hojjatabdollahi:
I think everybody agrees this makes more sense:
image

Are we waiting for confirmation from UI/UX? (since I don't think the original design was sanctioned by UI/UX to begin with)

wash2 (dismissed):

jacobgkau:
We're running it by Maria in Mattermost just in case. I do think Jonah's confusion was valid; when the option's off, it looks a little odd having space reserved for the checkmark but no empty box or anything to indicate why that space is reserved.

hojjatabdollahi:
We changed the order of the items in the context menu and changed the text for 2 items.

"Toggle maximized" => "Maximize or restore pane" [I reused the same token in the i18n file]

"Show header" => "Window header" [I kept the old token, and added another one since the previous one was an action, and the translations probably won't match].


I don't know the rule here regarding changing the wording. Should I have deleted the old "Show header" in all translation files?

jklgrasso:
I am not sure what to do there either. `show-headerbar` would no longer be used, correct? If so, it should be fine to remove it. ~~It would probably be best to ask in the same Mattermost thread.~~ Asked in Mattermost.

jacobgkau:
> I don't know the rule here regarding changing the wording. Should I have deleted the old "Show header" in all translation files?

I think we only want to touch the English translation file, and let Weblate handle the rest.

hojjatabdollahi:
Actually, I think we're good. 
`show-header` is still in use in the settings, so no need to delete it from the translation files.

image

jklgrasso (approved):
Context menus work as expected along with appearance and font settings, pane splits, copy/paste, tabs, and tab management (cycling left -> right, right -> left, opening, closing, etc.).

